PR #5, opened 2021-02-08
added support for EP0099

cadusk:
EP0099 is a stackable 4-channel relay hat that communicates with
platforms via i2c interface.

Manufacturer: 52pi
Wiki page: https://wiki.52pi.com/index.php/DockerPi_4_Channel_Relay_SKU:_EP-0099

### DIFF
--- a/ep0099/doc.go
+++ b/ep0099/doc.go
@@ -1,0 +1,12 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that caan be found in the LICENSE file.
+
+// Package ep0099
+//
+// Datasheet
+// https://wiki.52pi.com/index.php/DockerPi_4_Channel_Relay_SKU:_EP-0099
+//
+// Examples
+// https://github.com/geeekpi/dockerpi
+package ep0099
--- a/ep0099/doc.go
+++ b/ep0099/doc.go
@@ -1,12 +1,9 @@
-// Copyright 2018 The Periph Authors. All rights reserved.
+// Copyright 2021 The Periph Authors. All rights reserved.
 // Use of this source code is governed under the Apache License, Version 2.0
 // that caan be found in the LICENSE file.
 
-// Package ep0099
+// Package ep0099 controls a EP-0099 Raspberry Pi HAT with 4 relays via I2C.
 //
 // Datasheet
 // https://wiki.52pi.com/index.php/DockerPi_4_Channel_Relay_SKU:_EP-0099
-//
-// Examples
-// https://github.com/geeekpi/dockerpi
 package ep0099
--- a/ep0099/ep0099.go
+++ b/ep0099/ep0099.go
@@ -1,4 +1,4 @@
-// Copyright 2018 The Periph Authors. All rights reserved.
+// Copyright 2021 The Periph Authors. All rights reserved.
 // Use of this source code is governed under the Apache License, Version 2.0
 // that can be found in the LICENSE file.
 
@@ -10,14 +10,14 @@ import (
 	"periph.io/x/conn/v3/i2c"
 )
 
-var InvalidAddressError = errors.New("Invalid EP-0099 address")
-var InvalidChannelError = errors.New("Invalid EP-0099 channel")
+var errInvalidAddress = errors.New("Invalid EP-0099 address")
+var errInvalidChannel = errors.New("Invalid EP-0099 channel")
 
 type State byte
 
 const (
 	StateOff State = 0x00
-	StateOn  State = 0x01
+	StateOn  State = 0xFF
 )
 
 type Dev struct {
@@ -107,12 +107,12 @@ func isValidAddress(address uint16) error {
 		}
 	}
 
-	return InvalidAddressError
+	return errInvalidAddress
 }
 
 func (d *Dev) isValidChannel(channel uint8) error {
 	if _, exists := d.state[channel]; !exists {
-		return InvalidChannelError
+		return errInvalidChannel
 	}
 	return nil
 }
--- a/ep0099/ep0099.go
+++ b/ep0099/ep0099.go
@@ -87,7 +87,10 @@ func (s State) String() string {
 // Reset resets the registers to the default values.
 func (d *Dev) reset() error {
 	for channel := range d.state {
-		d.Off(channel)
+		err := d.Off(channel)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/ep0099/ep0099.go
+++ b/ep0099/ep0099.go
@@ -1,0 +1,129 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package ep0099
+
+import (
+	"errors"
+
+	"periph.io/x/conn/v3/i2c"
+)
+
+var InvalidAddressError = errors.New("Invalid EP-0099 address")
+var InvalidChannelError = errors.New("Invalid EP-0099 channel")
+
+type State byte
+
+const (
+	StateOff State = 0x00
+	StateOn  State = 0x01
+)
+
+type Dev struct {
+	i2c   i2c.Dev
+	state map[uint8]State
+}
+
+func New(bus i2c.Bus, address uint16) (*Dev, error) {
+	if err := isValidAddress(address); err != nil {
+		return nil, err
+	}
+
+	d := &Dev{
+		i2c:   i2c.Dev{Bus: bus, Addr: address},
+		state: buildChannelsList(),
+	}
+
+	if err := d.reset(); err != nil {
+		return nil, err
+	}
+
+	return d, nil
+}
+
+func (d *Dev) Halt() error {
+	return d.reset()
+}
+
+func (d *Dev) On(channel uint8) error {
+	if err := d.isValidChannel(channel); err != nil {
+		return err
+	}
+
+	_, err := d.i2c.Write([]byte{channel, byte(StateOn)})
+	d.state[channel] = StateOn
+	return err
+}
+
+func (d *Dev) Off(channel uint8) error {
+	if err := d.isValidChannel(channel); err != nil {
+		return err
+	}
+
+	_, err := d.i2c.Write([]byte{channel, byte(StateOff)})
+	d.state[channel] = StateOff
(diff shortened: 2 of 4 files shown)


ep0099/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel, line 1: Please update

ep0099/doc.go
```diff
@@ -0,0 +1,12 @@
...
+// Use of this source code is governed under the Apache License, Version 2.0
+// that caan be found in the LICENSE file.
+
+// Package ep0099
+//
+// Datasheet
+// https://wiki.52pi.com/index.php/DockerPi_4_Channel_Relay_SKU:_EP-0099
+//
+// Examples
```
maruel, line 10: Remove, not relevant for users not implementers.

ep0099/ep0099.go
```diff
@@ -0,0 +1,129 @@
...
+
+package ep0099
+
+import (
+	"errors"
+
+	"periph.io/x/conn/v3/i2c"
+)
+
+var InvalidAddressError = errors.New("Invalid EP-0099 address")
```
maruel, line 13: invalid (lower case)

ep0099/ep0099.go
```diff
@@ -0,0 +1,129 @@
...
+var InvalidChannelError = errors.New("Invalid EP-0099 channel")
+
+type State byte
```
maruel, line 16: Document
Why not a bool?

ep0099/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that caan be found in the LICENSE file.
+
+// Package ep0099
```
maruel, line 5: Package ep0099 controls a EP-0099 Raspberry Pi HAT with 4 relays via I²C.

or something like that.

ep0099/doc.go
```diff
@@ -0,0 +1,12 @@
...
+//
+// Datasheet
```
maruel, line 7: Can you tell which chip is being used. I almost see it at https://wiki.52pi.com/images/2/2e/Ddl-4.jpg but

ep0099/doc.go
cadusk: I based this implementation in a simple example I found in SN3218.
updating to 2021.

ep0099/doc.go
cadusk: looks like all other devices have links to data sheets in doc.go as I was trying to be consistent.

ep0099/ep0099.go
cadusk: after analyzing github actions build go-lint output, it suggests that this is renamed to "ErrInvalidAddress".
I will change it to that and make it lower case so it's internal to the package.

ep0099/ep0099.go
cadusk: the idea was to have StateOn and StateOff match the value that would be send to the device that represents the desired state so it is easier to read tests when the assertion to `dev.On` checks if the value recorded was StateOn.

it can be easily updated to bool and Write send 0x00 or 0xFF to the device.

ep0099/doc.go
cadusk: oops, yeah. my bad.

ep0099/doc.go
cadusk: The closest I could get was at https://wiki.52pi.com/images/5/52/Ddl-11.jpg

Apparently it reads "32F030F4P6".
https://www.st.com/en/microcontrollers-microprocessors/stm32f030f4.html#cad-resources

ep0099/doc.go
maruel: I was commenting on the Examples, not Datasheet.

ep0099/doc.go
maruel: Thanks.

ep0099/ep0099.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+	return nil
+}
+
+// Addresses in EP0099 are configured via DIP Switches on the board.
+// Up to 4 HATs can be stacked and each one need a different address to
+// work.
+func isValidAddress(address uint16) error {
+	validAddresses := [...]uint16{0x10, 0x11, 0x12, 0x13}
```
maruel, line 102: It's quite inefficient. Instead do:

```
switch address {
  case 0x10, 0x11, 0x12, x013:
    return nil
  default:
    return errors
}
```

ep0099/ep0099.go
```diff
@@ -0,0 +1,132 @@
...
+func (d *Dev) isValidChannel(channel uint8) error {
+	if _, exists := d.state[channel]; !exists {
+		return errInvalidChannel
+	}
+	return nil
+}
+
+// EP-0099 offers 4 channels per board
+func buildChannelsList() map[uint8]State {
+	// Using a map instead of list since indexes of channels are not zero-based
```
maruel, line 122: Nit: I'd prefer a [4] and just do -1 on the channel number. You'd save a memory allocations and enable better compiler checks.

ep0099/ep0099.go
```diff
@@ -0,0 +1,132 @@
...
+		if address == addr {
+			return nil
+		}
+	}
+
+	return errInvalidAddress
+}
+
+func (d *Dev) isValidChannel(channel uint8) error {
+	if _, exists := d.state[channel]; !exists {
```
maruel, line 114: why not: `return channel >= 1 && channel <= 4` ?

ep0099/ep0099.go
```diff
@@ -0,0 +1,132 @@
...
+func (s State) String() string {
+	if s == StateOff {
+		return "off"
+	}
+	return "on"
+}
+
+// Reset resets the registers to the default values.
+func (d *Dev) reset() error {
+	for channel := range d.state {
```
maruel, line 89: Note that this is operating in a non-deterministic order. This can be surprising.

ep0099/ep0099.go
cadusk: yeah... trying to make this extensible but... you're right. the device form factor will not change.

ep0099/ep0099.go
cadusk: ok

ep0099/ep0099.go
cadusk: ok

ep0099/ep0099.go
```diff
@@ -32,7 +32,7 @@ func New(bus i2c.Bus, address uint16) (*Dev, error) {
 
 	d := &Dev{
 		i2c:   i2c.Dev{Bus: bus, Addr: address},
-		state: buildChannelsList(),
+		state: [4]State{StateOff, StateOff, StateOff, StateOff},
```
maruel, line 35: It's the default so you can skip the whole line.

ep0099/ep0099.go
```diff
@@ -99,34 +98,14 @@ func (d *Dev) reset() error {
...
-	// values. That would cause loops to have to correct channel ids while
-	// looping through items or reading/setting values.
-	// With maps, keys correspond to actual channels on the board.
-	return map[uint8]State{
-		1: StateOff,
-		2: StateOff,
-		3: StateOff,
-		4: StateOff,
-	}
+func (d *Dev) isValidChannel(channel uint8) bool {
```
maruel, line 109: Doesn't need to be a method.

ep0099/ep0099.go
maruel: On one hand I would have wanted to use gpio.Level but at the same time it doesn't mean exactly the same thing; gpio.High doesn't necessarily mean "On".

The vision is that Dev could embed 4 Relay objects, which then would implement gpio.PinOut https://periph.io/x/conn/v3/gpio#PinOut

The value proposition is that then they are simply pins that can be forwarded to other systems.

pseudo code would look like

```
type Dev struct {
  Relays [4]Relay
}

type Relay struct {
  d*Dev   <- pointer back to the main struct
  state bool
}

type (r *Relay) Out(l gpio.Level) error {
	_, err := d.i2c.Write([]byte{channel, byte(StateOn)})
	r.state = StateOn
	return err
}
```

Then users can do

```
d := ep0099.New(...)
d.Relays[0].Out(gpio.High)
d.Relays[0].Out(gpio.Low)
fmt.Printf("%s\n", d.Relay[0].Read())
```

I don't know if users would see this as good ergonomics. The nice thing is that Relays is an array so the compiler enforces boundary checks for you. I try to make things generic when possible but it doesn't always make sense.

TL;DR: as you see fit.

maruel (approved):
I'm fine with committing as is or with optional changes, as you prefer.

ep0099/ep0099.go
cadusk: this makes sense. will remove it.

ep0099/ep0099.go
cadusk: Unless you're seeing a big toll on performance, I'd say it makes sense to have it centralized in a method than replicate the validation in multiple places.

cadusk:
I like this approach, I just don't see how to figure out which channel we're referring to in method `Out` since we're passing the index to the array, not to `dev`.

```golang
type (r *Relay) Out(l gpio.Level) error {
	_, err := d.i2c.Write([]byte{channel, byte(StateOn)})
	r.state = StateOn
	return err
}```

I've been learning a lot with this PR, maybe there's a way to figure out what channel would be without looping through Relays and comparing d to items in the array.

Maybe this solution needs some more thinking.
I am happy with the results so far and am ok with merging it like this.

File: ep0099/ep0099.go (hunk 12 above)
Comment (maruel) [reply ep0099/ep0099.go:108]: Oh, I meant it can be a stand alone function, it doesn't need a pointer receiver. What I meant is that this can be rewritten as:

```
func isValidChannel(channel uint8) bool {
  return channel >= 1 && channel <= 4
}
```

It's not material, so I'll commit and this can be changed later if you want.

maruel:
One last point: do you mind following the instruction at https://github.com/periph/devices/pull/5#issuecomment-775185342?

ep0099/ep0099.go
cadusk: oh, I get it now.
it really makes sense.

maruel:
gohci

maruel:
Thanks!

PR #7, opened 2021-03-08
Prevent WaitForEdge() returning early after first call to selectCard()

brackendawson:
Calling LowLevel.Init() in selectCard() clears the device interrupt
as a side effect, but another interrupt will occur due to the
subsquent operations. This additional host interrupt must be cleared
before the next call to selectCard.

https://github.com/google/periph/issues/425


brackendawson:
@googlebot I signed it!

maruel:
Thanks!

maruel:
Ping me on slack to remind me to make a release because I don't have time today and I know I'll forget.

PR #8, opened 2021-03-20
waveshare2in13v2: add example

kortschak:
This is a direct adaptation of the epd package example.

Please take a look.

### DIFF
--- a/waveshare2in13v2/example_test.go
+++ b/waveshare2in13v2/example_test.go
@@ -1,0 +1,120 @@
+// Copyright 2021 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package waveshare2in13v2_test
+
+import (
+	"image"
+	"image/draw"
+	"log"
+
+	"golang.org/x/image/font"
+	"golang.org/x/image/font/basicfont"
+	"golang.org/x/image/math/fixed"
+
+	"github.com/fogleman/gg"
+	"github.com/golang/freetype/truetype"
+	"golang.org/x/image/font/gofont/goregular"
+
+	"periph.io/x/conn/v3/spi/spireg"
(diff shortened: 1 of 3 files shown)


belak:
Nice! I think this is exactly what I did when testing mine.

waveshare2in13v2/example_test.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2018 The Periph Authors. All rights reserved.
```
maruel, line 1: 2021

waveshare2in13v2/example_test.go
```diff
@@ -0,0 +1,121 @@
...
+	defer b.Close()
+
+	dev, err := waveshare2in13v2.NewHat(b, &waveshare2in13v2.EPD2in13v2) // Display config and size
+	if err != nil {
+		log.Fatalf("failed to initialize epd: %v", err)
+	}
+
+	// Draw on it. Black text on a white background.
+	img := image1bit.NewVerticalLSB(dev.Bounds())
+	// Note: this code is commented out so periph does not depend on:
```
maruel, line 37: This is not a problem now that this package uses go modules. So we should revisit these examples and use golang.org/x/image when it makes sense. So feel free to uncomment if you want.

maruel (approved):
looks good, please update the year, and you can uncomment the code if you want. If you do, make sure updated go.mod and go.sum are included.

waveshare2in13v2/example_test.go
kortschak: I left 2018 since the code largely comes from a file dated 2018, but I will change it.

waveshare2in13v2/example_test.go
kortschak: Done

kortschak:
PTAL

maruel:
Thanks!

PR #9, opened 2021-03-20
[TLV493D] Method to calibrate temperature offset

bezineb5:
Hello,

This is an addition to the TLV493D driver, in order to calibrate the "temperature offset". According to the [documentation](https://www.infineon.com/dgdl/Infineon-TLV493D-A1B6_3DMagnetic-UM-v01_03-EN.pdf?fileId=5546d46261d5e6820161e75721903ddd), section 3.2 "Calculation of the temperature", there is an Offset compensation, with a note: "The 340 as typical value may vary from device to device. Therefore for an accurate measurement the value at 25°C needs to be readout first and used as (correction) value instead of 340."

So this PR is basically the implementation of this.

Thanks!
Benjamin

### DIFF
--- a/tlv493d/tlv493d.go
+++ b/tlv493d/tlv493d.go
@@ -30,6 +30,9 @@ const (
 	HighPrecisionWithTemperature Precision = 0
 	// LowPrecision reads only 8-bits for each axis. Temperature is not read.
 	LowPrecision Precision = 1
+	// DefaultTemperatureOffsetCompensation is the temperature offset compensation used by default, following the documentation advices.
+	// For accurate results, please calibrate it using CalibrateTemperatureOffsetCompensation method.
+	DefaultTemperatureOffsetCompensation = 340
 )
 
 const (
@@ -137,7 +140,7 @@ var DefaultOpts = Opts{
 	EnableTemperatureMeasurement:  true,
 	InterruptPadEnabled:           false,
 	ParityTestEnabled:             true,
-	TemperatureOffsetCompensation: 340, // As per the documentation, can be calibrated for better precision
+	TemperatureOffsetCompensation: DefaultTemperatureOffsetCompensation, // As per the documentation, can be calibrated for better precision
 }
 
 // Sample contains the metrics measured by the sensor
@@ -420,6 +423,18 @@ func (d *Dev) StopContinousRead() {
 	d.stop <- struct{}{}
 	d.stop = nil
 	d.continuousReadWG.Wait()
+}
+
+// CalibrateTemperatureOffsetCompensation computes the temperature offset compensation based upon a reading from the sensor compared to the actual temperature.
+// The returned value must be stored and passed in Opts.TemperatureOffsetCompensation in future uses of TLV493D driver.
+// See TLV493D user manual, "3.2 Calculation of the temperature" for more information
+func CalibrateTemperatureOffsetCompensation(temperatureOffsetCompensation int, measuredTemperature physic.Temperature, actualTemperature physic.Temperature) int {
+	// Compute measurement based upon reference documentation
+	rawTemp := int((measuredTemperature-referenceTemperature)/temperatureScaling) + temperatureOffsetCompensation
+	// Compute the new offset by matching the raw measurement between the measure and the actual value
+	newOffset := rawTemp - int((actualTemperature-referenceTemperature)/temperatureScaling)
+
+	return newOffset
 }
 
 func bestModeForFrequency(frequency physic.Frequency) (Mode, error) {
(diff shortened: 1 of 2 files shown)


bezineb5:
@googlebot I fixed it.

tlv493d/tlv493d.go
```diff
@@ -422,6 +422,18 @@ func (d *Dev) StopContinousRead() {
 	d.continuousReadWG.Wait()
 }
 
+// CalibrateTemperature sets the temperature offset compensation based upon a reading from the sensor compared to the actual temperature
```
maruel, line 425: Since it should be set in Opts.TemperatureOffsetCompensation, I think it could be a stand alone function. This way people could calculate the value, then save it locally so it is specified in Opts, instead of changing the value once the device is initialized.

What do you think?

bezineb5: Yes, I agree, however:
* the new value of Opts.TemperatureOffsetCompensation is returned as output of the method - but that's probably not explicit in the doc
* in order to extract the raw reading from the measured temperature, I need to use the Opts.TemperatureOffsetCompensation used for computing it, that's why I set this as a member function.

Basically, the function needs access to the raw value returned by the sensor and the matching temperature. So, what I can propose:
1. Make it more explicit in the documentation that the output of the function must be saved for reuse and put in Opts.TemperatureOffsetCompensation next time.
2. Make the function standalone and ask for Opts.TemperatureOffsetCompensation as input

I agree on doing 2, it makes sense. I'll do a PR shortly with it.

Thanks for the feedback!

tlv493d/tlv493d.go
```diff
@@ -422,6 +425,18 @@ func (d *Dev) StopContinousRead() {
...
 }
 
+// CalibrateTemperatureOffsetCompensation computes the temperature offset compensation based upon a reading from the sensor compared to the actual temperature.
+// The returned value must be stored and passed in Opts.TemperatureOffsetCompensation in future uses of TLV493D driver.
+// See TLV493D user manual, "3.2 Calculation of the temperature" for more information
+func CalibrateTemperatureOffsetCompensation(temperatureOffsetCompensation int, measuredTemperature physic.Temperature, actualTemperature physic.Temperature) int {
+	// Compute measurement based upon reference documentation
+	rawTemp := int((measuredTemperature-referenceTemperature)/temperatureScaling) + temperatureOffsetCompensation
+	// Compute the new offset by matching the raw measurement between the measure and the actual value
+	newOffset := rawTemp - int((actualTemperature-referenceTemperature)/temperatureScaling)
```
maruel, line 435: you could return it directly.

tlv493d/tlv493d_test.go
```diff
@@ -326,6 +326,17 @@ func TestTLV493D_Configuration(t *testing.T) {
 	}
 }
 
+func Test_TLV493D_TemperatureCalibration(t *testing.T) {
```
maruel, line 329: Thanks for the unit test!

tlv493d/tlv493d.go
```diff
@@ -422,6 +425,18 @@ func (d *Dev) StopContinousRead() {
 	d.continuousReadWG.Wait()
 }
 
+// CalibrateTemperatureOffsetCompensation computes the temperature offset compensation based upon a reading from the sensor compared to the actual temperature.
+// The returned value must be stored and passed in Opts.TemperatureOffsetCompensation in future uses of TLV493D driver.
+// See TLV493D user manual, "3.2 Calculation of the temperature" for more information
```
maruel, line 430: Generally prefer to end sentences with a period.

tlv493d/tlv493d.go
```diff
@@ -422,6 +425,18 @@ func (d *Dev) StopContinousRead() {
 	d.continuousReadWG.Wait()
 }
 
+// CalibrateTemperatureOffsetCompensation computes the temperature offset compensation based upon a reading from the sensor compared to the actual temperature.
+// The returned value must be stored and passed in Opts.TemperatureOffsetCompensation in future uses of TLV493D driver.
```
maruel, line 429: The doc style is to keep an empty line after the describing sentence. Not a big deal.

tlv493d/tlv493d.go
maruel: Great idea.

maruel:
Committing since it's good as is. A few nits if you want to make a follow up but no big deal either way.

tlv493d/tlv493d.go
bezineb5: Fixed

tlv493d/tlv493d.go
bezineb5: Fixed

tlv493d/tlv493d_test.go
bezineb5: You're welcome!

tlv493d/tlv493d.go
bezineb5: Fixed

PR #10, opened 2021-03-21
[TLV493D] Method to calibrate temperature offset

bezineb5:
Fixes from previous PR.
Sorry, I didn't have time to update the previous PR with your comments, so here they are. I prefer to finish now otherwise I'll forget.


maruel:
Thanks!

Issue #11, opened 2021-04-04
Adding support for Inky auto-detecting board type


Thereatra:
I have been using the Inky device driver for a project and needed the automatic board detection feature of the official Python library. There is no official board documentation so I have simply ported the feature over, from the Python library, for my own project. The code is still rather crude but I would be interested in cleaning it up and adding it to the device driver.
Where should I start?

maruel:
Hi! I don't have such devices so I can't help directly.
According to history, you may want to talk with @davidsansome and @hatstand for guidance.

davidsansome:
That sounds cool, I'd be happy to review a pull request for board detection! I've got a red/black wHAT I can try it with, and I think @hatstand has some kind of pHAT.

I'm not sure if there's a convention for hardware detection methods elsewhere in Periph, but what about a function in inky.go that returns an Opts struct?

Thereatra:
I'm new to both the hardware and go space so I'm not that sure of the conventions, but since the Inky is designed as a Pi hat out of the box I was thinking a function that could return a Dev object with the defaults for pi, would be very useful. This is also what I have implemented currently, but stripping it down if it is out of scope for Periph shouldn't be a problem.

maruel:
Thanks @Thereatra for doing the change! I still need to tag a new version, will do soon.

PR #12, opened 2021-04-05
[Inky] autodetect board opts and pHAT v2 support

Thereatra:
I have added auto opts detect and dev struct creationg with defaults from the detected opts as mentioned in #11.

While I was at it I also added support for pHAT v2 which is simply a pHAT with a [small resolution difference](https://github.com/pimoroni/inky/blob/85f97ffa9ae775d4de71d5179af792ccb7080ee0/library/inky/phat.py#L15).

### DIFF
--- a/inky/inky.go
+++ b/inky/inky.go
@@ -9,14 +9,13 @@ import (
 	"fmt"
 	"image"
 	"image/color"
-	"periph.io/x/conn/v3/gpio/gpioreg"
-	"periph.io/x/conn/v3/i2c/i2creg"
-	"periph.io/x/conn/v3/spi/spireg"
 	"time"
 
 	"periph.io/x/conn/v3"
 	"periph.io/x/conn/v3/display"
 	"periph.io/x/conn/v3/gpio"
+	"periph.io/x/conn/v3/gpio/gpioreg"
+	"periph.io/x/conn/v3/i2c"
 	"periph.io/x/conn/v3/physic"
 	"periph.io/x/conn/v3/spi"
 )
@@ -114,10 +113,10 @@ type Opts struct {
 	BorderColor Color
 }
 
-// DetectOpts tries to read the device opts from EEPROM, it is recommended to use I2C bus "1"
-func DetectOpts(i2cBus string) (*Opts, error) {
+// DetectOpts tries to read the device opts from EEPROM, it is recommended to use I2C bus '1'.
+func DetectOpts(bus i2c.Bus) (*Opts, error) {
 	// Read data from EEPROM
-	data, err := readEep(i2cBus)
+	data, err := readEep(bus)
 	if err != nil {
 		return nil, fmt.Errorf("failed to detect Inky board: %v", err)
 	}
@@ -204,14 +203,10 @@ func New(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, o *Opts
 	return d, nil
 }
 
-// NewDetected tries to open a handle to an Inky pHat or wHAT automatically detected from EEPROM
-func NewDetected() (*Dev, error) {
-	opts, err := DetectOpts("1")
-	if err != nil {
-		return nil, err
-	}
-
-	spiPort, err := spireg.Open("SPI0.0")
+// NewDetected tries to open a handle to an Inky pHat or wHAT automatically detected from EEPROM.
+// For a Raspberry Pi running linux the SPI port will usually be 'SPI0.0' and I2C bus '1'.
+func NewDetected(port spi.Port, bus i2c.Bus) (*Dev, error) {
+	opts, err := DetectOpts(bus)
 	if err != nil {
 		return nil, err
 	}
@@ -220,7 +215,7 @@ func NewDetected() (*Dev, error) {
 	resetPin := gpioreg.ByName("27")
 	busyPin := gpioreg.ByName("17")
 
-	return New(spiPort, dcPin, resetPin, busyPin, opts)
+	return New(port, dcPin, resetPin, busyPin, opts)
 }
 
 // Dev is a handle to an Inky.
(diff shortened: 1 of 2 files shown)


Thereatra:
@googlebot I signed it!

inky/inky.go
```diff
@@ -9,6 +9,9 @@ import (
 	"fmt"
 	"image"
 	"image/color"
+	"periph.io/x/conn/v3/gpio/gpioreg"
```
maruel, line 12: Please setup goimports in your IDE.

inky/inky.go
```diff
@@ -106,6 +114,44 @@ type Opts struct {
 	BorderColor Color
 }
 
+// DetectOpts tries to read the device opts from EEPROM, it is recommended to use I2C bus "1"
+func DetectOpts(i2cBus string) (*Opts, error) {
```
maruel, line 118: This should accept an already open bus.

inky/inky.go
```diff
@@ -148,13 +194,35 @@ func New(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, o *Opts
...
 }
 
+// NewDetected tries to open a handle to an Inky pHat or wHAT automatically detected from EEPROM
+func NewDetected() (*Dev, error) {
+	opts, err := DetectOpts("1")
+	if err != nil {
+		return nil, err
+	}
+
+	spiPort, err := spireg.Open("SPI0.0")
```
maruel, line 214: Do not hardcode spi ports, this can change wildly, e.g. someone connecting it over USB on Windows.

Thereatra:
I've changed it so `DetectOpts` and the related functions all use an open port instead.

I guess I just assumed that the display would always run on a raspberry pi with linux as it is built for. I have have removed the hardcoded SPI port as specified, however the `NewDetected` still relies on the assumption that the Inky is used as a pi hat with the specific gpio ports. I personally feel like this makes sense due to the nature of the board, but if it's out of scope/too specific for a device driver, then it's probably better to just remove the function rather than make it take the same parameters as `New` minus the opts param.

inky/inky.go
```diff
@@ -106,6 +113,44 @@ type Opts struct {
 	BorderColor Color
 }
 
+// DetectOpts tries to read the device opts from EEPROM, it is recommended to use I2C bus '1'.
```
maruel, line 116: > it is recommended to use I2C bus '1'.

I'd recommend to remove.

inky/inky.go
```diff
@@ -148,13 +193,31 @@ func New(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, o *Opts
...
+		d.bounds = image.Rect(0, 0, 122, 250)
+		d.flipVertically = true
 	case WHAT:
 		d.bounds = image.Rect(0, 0, 400, 300)
 	}
 
 	return d, nil
 }
 
+// NewDetected tries to open a handle to an Inky pHat or wHAT automatically detected from EEPROM.
```
maruel, line 206: Ok so eventually we could have New() be replaced to do the auto detection?

Is the I²C channel used for anything else? Is it worth forcing users to use it all the times?

inky/inky.go
```diff
@@ -148,13 +193,31 @@ func New(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinIn, o *Opts
...
+// For a Raspberry Pi running linux the SPI port will usually be 'SPI0.0' and I2C bus '1'.
+func NewDetected(port spi.Port, bus i2c.Bus) (*Dev, error) {
```
maruel, line 208: I think I'd like to have either DetectOpts() or NewDetected() exported but not both. I'm edging on DetectOpts since it's more composable and removes the dependency on gpioreg.

maruel (commented):
Thanks it's good. Just have a question about the API surface.

Sorry for the delay, a bit busy over here.

inky/inky.go
Thereatra: That's a fair point, I was just pulling it from the Periph Raspberry Pi setup recommendations, but I can see how that should probably be kept out of function documentation

inky/inky.go
Thereatra: For general use it should be able to replace New() but replacing it completely would mean a user would be unable to use an Inky with faulty EEPROM, due to only being able to initialize it through detection. From reading the official python code faulty EEPROM doesn't seem to affect Inky performance, and from personal experience the python library has failed to detect one of my Inky, so this does not seem like that unlikely of a scenario.

The I²C channel is only used for reading the EEPROM, which is needed in order to detect the model. I could make NewDetected() handle opening the bus itself, but that would require using a predetermined bus and I got impression that that would be a bad approach?

inky/inky.go
Thereatra: As a user I'd personally prefer both, DetectOpts() for flexibility and NewDetected() ease of access, but I can understand why only having one of them exported would be preferable. In that case I think it would be better to keep DetectOps(), for the reasons you have mentioned as well as the ability to still detect the Inky model when using a non standard pinout on the device.

inky/inky.go
maruel: If bad EEPROM is a known failure mode, then I prefer the users to have to explicitly always call DetectOpts() (or DetectDevice() ?) then call New() themselves.

Thereatra: I don't know if bad EEPROM is a known failure, this was purely anecdotal. My point was just that it could happen, so the library should probably allow manual new instances of Dev to be created. 
I am not really sure how to interpret your comment in terms of code changes, both in relation to which functions should be available and how to handle the I²C channel parameter.

maruel: Let me clarify what I mean. My expectation is to use this flow:

```go
opts, err := inky.DetectOpts(bus)
if err != nil {
  // Use hardcoded default values.
  opts = inky.Opts{...}
}
d, err := inky.New(port, dc, reset, busy, opts)
if err != nil {
  log.Fatal(err)
}
// use d
```

So NewDetected is unnecessary.

Thereatra: Yeah okay, sounds good I'll get on it. I think that is what I was trying to express in the conversation below this one

maruel:
Thanks! Apologies for the slow back and forth.

One potential follow up: I think it would be interesting to add an example about how to leverage the new function in https://github.com/periph/devices/blob/main/inky/example_test.go.

PR #13, opened 2021-04-27
ds248x.go is modified to support DS2482-800 in addition to DS2482-100. ds18b20.go is modified to allow 2 separate calls for conversion start and result read out.

audrius-paskevicius:


maruel:
Please fix the checks, gofmt, misspell, test crashing.

maruel:
You should be able to reproduce locally with "go test"

ds248x/ds248x.go
```diff
@@ -160,6 +161,7 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
 	return onewire.Search(d, alarmOnly)
 }
 
+// ChannelSelect ...
```
maruel, line 164: ?

maruel: I meant the "..." this is not a proper documentation.

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+	}
+	return
+}
+
+// SelectedChannel function is to read with 1-w channel selected on DS2482-800. 
+// On other chips it always returns 0. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+//
+// On error returns 255.
+//
```
maruel, line 200: Remove.

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+
+	}
+	return
+}
+
+// SelectedChannel function is to read with 1-w channel selected on DS2482-800. 
+// On other chips it always returns 0. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+//
+// On error returns 255.
```
maruel, line 199: Why 255? It seems a bit random since it's a int, not a uint8.

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+//
+func (d *Dev) SelectedChannel() (ch int) {
```
maruel, line 201: Do not use named return value unless necessary.

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
 	return onewire.Search(d, alarmOnly)
 }
 
+// ChannelSelect function is for selecting one of eight 1-w channels on DS2482-800. 
+// On other chips it does nothing. Function silently limits channel selection between
+// 0 and 7. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+//
+// Communication error is returned if present.
+//
```
maruel, line 170: Remove empty //

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+func (d *Dev) ChannelSelect(ch int) (err error) {
+	switch d.isDS248x {
+	case isDS2482x100:
+
```
maruel, line 174: return nil if ch == 0, otherwise an error? Same for isDS2483.

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+	case isDS2482x800:
+		if ch < 0 {
+			ch = 0
+		}
+		if ch > 7 {
+			ch = 7
+		}
+		csc := []byte{cscIO0w, cscIO1w, cscIO2w, cscIO3w, cscIO4w, cscIO5w, cscIO6w, cscIO7w}
```
maruel, line 182: Make it a global, otherwise it'll be constructed at each call.

ds18b20/ds18b20.go
```diff
@@ -33,6 +33,17 @@ func ConvertAll(o onewire.Bus, maxResolutionBits int) error {
 	return nil
 }
 
+// StartAll starts a conversion on all DS18B20 devices on the bus.
```
maruel, line 36: FTR, I don't understand ConvertAll() either, since the resolution cannot be set, I don't understand the purpose of this function.

ds248x/ds248x.go
```diff
@@ -154,6 +161,70 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+	switch d.isDS248x {
+	case isDS2482x100:
+
+	case isDS2482x800:
+		var sch [1]byte
+		if err := d.i2c.Tx([]byte{cmdSetReadPtr, regCSR}, sch[:]); err != nil {
+			ch = 255
+			return 
+		}
+		csc := []byte{cscIO0r, cscIO1r, cscIO2r, cscIO3r, cscIO4r, cscIO5r, cscIO6r, cscIO7r}
```
maruel, line 212: Move to a global, using [...]byte{} so it's an array, not a slice.

maruel (commented):
Now getting to the more substantive comments. Also please make the checks green by running gofmt. Popular IDEs will do it automatically for you.

ds18b20/ds18b20.go
audrius-paskevicius: Basically it's for savings of bandwidth. Settings are retained inside ds18b20. There is command to start conversion on all devices on the bus (can be up to 64) at once. What I usually do is: after issuing conversion start command I'm leaving the bus and returning later to collect data. When all data is collected on this particular bus I send conversion start command again.

ds248x/ds248x.go
audrius-paskevicius: Sorry it's a habit from memory restrained machines.

ds248x/ds248x.go
```diff
@@ -315,18 +392,44 @@ var sleep = time.Sleep
...
+	cscIO4w = 0xB4 // channel 4 writing
+	cscIO4r = 0x9C // channel 4 reading
+	cscIO5w = 0xA5 // channel 5 writing
+	cscIO5r = 0x95 // channel 5 reading
+	cscIO6w = 0x96 // channel 6 writing
+	cscIO6r = 0x8E // channel 6 reading
+	cscIO7w = 0x87 // channel 7 writing
+	cscIO7r = 0x87 // channel 7 reading
+
+	isDS2482x100 = 0 // DS2482-100 selected
```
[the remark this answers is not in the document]
audrius-paskevicius: That's a good idea.

ds248x/ds248x_test.go
```diff
@@ -11,14 +11,18 @@ import (
 	"periph.io/x/conn/v3/i2c/i2ctest"
 )
 
+// TestNew sadly is testing only for DS2483.
```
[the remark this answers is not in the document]
audrius-paskevicius: Problem is on I couldn't get these synthetic tests to branch. Chip is  identified by probing for some particular registers and branching on transmission results.

ds248x/ds248x.go
audrius-paskevicius: From 8 bit time... Could be more elegant by I wanted only one variable in return.

ds248x/ds248x.go
audrius-paskevicius: nil is returned by default. I'm a bit reluctant to produce too many errors. What wrong with doing nothing on these particular chips?

audrius-paskevicius:
I wonder about this gofmt error. I'm using Visual Studio Code. It does some checking and formatting. Maybe some setting mismatch? Problem is that I don't see on what line this test is failing.

ds18b20/ds18b20.go
audrius-paskevicius: ConvertAll resolution setting is just to determine how long conversion is going to take and wait for this time. Resolution is programmed into ds18b20 during initialisation at function New.

ds248x/ds248x.go
maruel: Silent failure is super hard to diagnose, and it's even harder if the user is not well versed in Go.

ds18b20/ds18b20.go
maruel: ok

ds248x/ds248x.go
```diff
@@ -15,6 +16,15 @@ import (
 	"periph.io/x/conn/v3/onewire"
 )
 
+// DS248xType for chip connected to the system identification (ds2482-100, ds2482-800, ds2483).
+type DS248xType uint8
+
+const (
+	isDS2482x100 = 0 // DS2482-100 selected
```
maruel, line 23: You need to call out explicitly with the time. I recommend using iota, which will define the rest of the typed values for you.

ds248x/ds248x.go
```diff
@@ -154,6 +170,73 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+	var err error = nil
+	switch d.isDS248x {
+	case isDS2482x100:
+		// if ch != 0 {
+		// 	err = errors.New("channel != 0")
+		// 	return fmt.Errorf("ds2482-100: error while selecting channel: %s", err)
+		// }
+	case isDS2482x800:
+		if ch < 0 {
+			err = errors.New("channel < 0")
```
maruel, line 188: I wouldn't bother with different error string for <0 or >7. Also why create two errors?

ds248x/ds248x.go
```diff
@@ -315,18 +404,41 @@ var sleep = time.Sleep
...
+	cscIO4r = 0x9C // channel 4 reading
+	cscIO5w = 0xA5 // channel 5 writing
+	cscIO5r = 0x95 // channel 5 reading
+	cscIO6w = 0x96 // channel 6 writing
+	cscIO6r = 0x8E // channel 6 reading
+	cscIO7w = 0x87 // channel 7 writing
+	cscIO7r = 0x87 // channel 7 reading
 )
+
+var cscw = []byte{cscIO0w, cscIO1w, cscIO2w, cscIO3w, cscIO4w, cscIO5w, cscIO6w, cscIO7w}
```
maruel, line 443: Use `[...]` so it's an array instead of a slice.
https://blog.golang.org/slices-intro

ds248x/ds248x.go
```diff
@@ -179,31 +179,23 @@ func (d *Dev) ChannelSelect(ch int) error {
 	var err error = nil
 	switch d.isDS248x {
 	case isDS2482x100:
-		// if ch != 0 {
-		// 	err = errors.New("channel != 0")
-		// 	return fmt.Errorf("ds2482-100: error while selecting channel: %s", err)
-		// }
+		err = errors.New("unsupported (ds2482-100 has only channel 0)")
```
maruel, line 182: Same here and for isDS2483; return errors.New("channel out of range").

ds248x/ds248x.go
```diff
@@ -440,5 +434,5 @@ const (
 	cscIO7r = 0x87 // channel 7 reading
 )
 
-var cscw = []byte{cscIO0w, cscIO1w, cscIO2w, cscIO3w, cscIO4w, cscIO5w, cscIO6w, cscIO7w}
+var cscw = [...]byte{cscIO0w, cscIO1w, cscIO2w, cscIO3w, cscIO4w, cscIO5w, cscIO6w, cscIO7w}
 var cscr = []byte{cscIO0r, cscIO1r, cscIO2r, cscIO3r, cscIO4r, cscIO5r, cscIO6r, cscIO7r}
```
maruel, line 438: Same here, `[...]`

maruel:
The check https://github.com/periph/devices/pull/13/checks?check_run_id=2501249442 still fails with:

```
Please run `gofmt -s -w` on the following files:

- ds248x/ds248x.go
```

ds248x/ds248x.go
audrius-paskevicius: For ds2482-100 and ds2483 errors.New("channel out of range") would be misinformation. There is no channel to choose! My suggestion is either explicitly tell user that, either silently do nothing. I prefer the second by cause for ds2482-100 and ds2483 communication would continue on same channel any way. Migration from ds2482-800 to ds2482-100 would be simpler.

audrius-paskevicius:
I did run `gofmt -s -w` on my RaspberryPi for ds248x.go.

ds248x/ds248x.go
```diff
@@ -15,6 +16,15 @@ import (
 	"periph.io/x/conn/v3/onewire"
 )
 
+// DS248xType for chip connected to the system identification (ds2482-100, ds2482-800, ds2483).
```
maruel, line 19: Please unexport it since it's not usable by the users.

ds248x/ds248x.go
```diff
@@ -154,6 +170,65 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+// ChannelSelect function is for selecting one of eight 1-w channels on DS2482-800.
+// On other chips it does nothing. Channel selection values are between
+// 0 and 7. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+// Communication error is returned if present.
+func (d *Dev) ChannelSelect(ch int) error {
+	var err error = nil
+	switch d.isDS248x {
+	case isDS2482x100:
+		// err = errors.New("unsupported (ds2482-100 has only channel 0)")
```
maruel, line 182: Remove the comments.

ds248x/ds248x.go
```diff
@@ -154,6 +170,65 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+		}
+		buf := []byte{cmdChannelSelect, cscw[ch]}
+		if err = d.i2c.Tx(buf, nil); err != nil {
+			return fmt.Errorf("ds2482-800: error while selecting channel: %s", err)
+		}
+	case isDS2483:
+		// err = errors.New("unsupported (ds2483 has only channel 0)")
+		// return fmt.Errorf("ds2483: error while selecting channel: %s", err)
+	default:
+		err = errors.New("wrong chip")
```
maruel, line 197: Inline

ds248x/ds248x.go
```diff
@@ -154,6 +170,65 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
 	return onewire.Search(d, alarmOnly)
 }
 
+// ChannelSelect function is for selecting one of eight 1-w channels on DS2482-800.
+// On other chips it does nothing. Channel selection values are between
+// 0 and 7. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+// Communication error is returned if present.
+func (d *Dev) ChannelSelect(ch int) error {
+	var err error = nil
```
maruel, line 179: This variable is unnecessary

ds248x/ds248x.go
```diff
@@ -154,6 +170,65 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+			return fmt.Errorf("ds2482-800: error while selecting channel: %s", err)
+		}
+	case isDS2483:
+		// err = errors.New("unsupported (ds2483 has only channel 0)")
+		// return fmt.Errorf("ds2483: error while selecting channel: %s", err)
+	default:
+		err = errors.New("wrong chip")
+		return fmt.Errorf("ds248x: error while selecting channel: %s", err)
+	}
+	return err
```
maruel, line 200: This will not be needed once you convert all the case to have a return statement.

ds248x/ds248x_test.go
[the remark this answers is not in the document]
maruel: One way to handle that is to create a i2ctest.Record, record a session then play it back in the unit test.

Or you can write it manually yourself by putting the right magic values, as you did in the comments.

I think it'd be a good idea to add this test.

ds248x/ds248x.go
audrius-paskevicius: Some environments complains about this.

ds18b20/ds18b20.go
```diff
@@ -33,6 +33,14 @@ func ConvertAll(o onewire.Bus, maxResolutionBits int) error {
...
+// Similar to ConvertAll but returns without waiting for converion to finish.
```
maruel, line 37: conversion

ds248x/ds248x.go
maruel: "invalid channel" then. That's generic yet factual. On isDS2482x100 and isDS2483 it should only accept 0. The code can look like:

```
case isDS2482x100, isDS2483:
  if ch == 0 {
    return errors.New("invalid channel")
  }
  return nil
```

ds248x/ds248x.go
maruel: It's because you don't return properly in the switch clauses.

ds248x/ds248x.go
```diff
@@ -154,6 +170,62 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+}
+
+// SelectedChannel function is to read with 1-w channel selected on DS2482-800.
+// On other chips it always returns 0. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+// On error returns 255.
+func (d *Dev) SelectedChannel() int {
+	ch := 0
+	switch d.isDS248x {
+	case isDS2482x100:
```
maruel, line 210:
```
case isDS2482x100, isDS2483:
  return 0
case isDS2482x800:
  ...
default:
  return 0
```

remove line 208, 226.

ds248x/ds248x.go
```diff
@@ -154,6 +170,55 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+// ChannelSelect function is for selecting one of eight 1-w channels on DS2482-800.
+// On other chips it does nothing. Channel selection values are between
+// 0 and 7. It is expected that application keeps track of
+// with 1-w device is connected to with channel.
+// Communication error is returned if present.
+func (d *Dev) ChannelSelect(ch int) error {
+	switch d.isDS248x {
+	case isDS2482x800:
+		if ch < 0 || ch > 7 {
+			return fmt.Errorf("%s: error while selecting channel: %s", d.String(), errors.New("channel out of range 0...7"))
```
maruel, line 182: There is no need to wrap an error into an error, replace the line with this;

```
return fmt.Errorf("%s: channel out of range 0...7", d)
```

ds248x/ds248x.go
```diff
@@ -154,6 +170,55 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+		}
+		buf := []byte{cmdChannelSelect, cscw[ch]}
+		if err := d.i2c.Tx(buf, nil); err != nil {
+			return fmt.Errorf("%s: error while selecting channel: %s", d.String(), err)
+		}
+		return nil
+	case isDS2482x100, isDS2483:
+		if ch != 0 {
+			return fmt.Errorf("%s: error while selecting channel: %s", d.String(), errors.New("invalid channel"))
```
maruel, line 191: Same.

ds248x/ds248x.go
```diff
@@ -154,6 +170,55 @@ func (d *Dev) Search(alarmOnly bool) ([]onewire.Address, error) {
...
+		}
+		return nil
+	default:
+		return fmt.Errorf("ds248x: error while selecting channel: %s", errors.New("wrong chip"))
```
maruel, line 195: Same.

ds248x/ds248x.go
maruel: One last change: I would prefer to change it to -1? This would remove an hardcoded value (255) that has no immediate meaning to the user. -1 is more often used as a sentinel value and is easier to check, like `if ch < 0 {`

maruel:
Thanks for the change!

maruel:
I'll try to do a new tag later today.

maruel:
Tagged as v3.6.10.

PR #14, opened 2021-05-08
ssd1306: Add SetDisplayStartLine

lutzky:
This is useful for controlled, smooth, vertical scrolling (to new content, not a looping scroll like `Dev.Scroll`. This is 10.1.6 from page 36 of [the datasheet].

With double-buffering, a 128x32 ssd1306 can scroll indefinitely. This can be done by setting `Opts.H: 64` (which does unfortunately dim the display a little bit), and using the lower 32 lines (which are off-screen) as a second image buffer.

[the datasheet]: https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf

### DIFF
--- a/ssd1306/ssd1306.go
+++ b/ssd1306/ssd1306.go
@@ -279,6 +279,17 @@ func (d *Dev) SetContrast(level byte) error {
 	return d.sendCommand([]byte{0x81, level})
 }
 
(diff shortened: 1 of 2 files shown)


maruel:
gohci

maruel:
Thanks! I'm waiting for another PR to be completed and I'll do a release.

PR #16, opened 2021-08-10
apa102: allow to specify SPI Mode in Opts

alexgth:
Some devices such as the secondary SPI Port of the Raspberry Pi 3 don't support the default SPI Mode 3.
This Change allows to set the SPI Mode to any other by changing the Opts.

### DIFF
--- a/apa102/apa102.go
+++ b/apa102/apa102.go
@@ -92,8 +92,12 @@ type Opts struct {
 // As per APA102-C spec, the chip's max refresh rate is 400hz.
 // https://en.wikipedia.org/wiki/Flicker_fusion_threshold is a recommended
 // reading.
-func New(p spi.Port, o *Opts) (*Dev, error) {
-	c, err := p.Connect(20*physic.MegaHertz, spi.Mode3, 8)
+//
+// Most devices can use spi.Mode3, but the raspberry pi 3 secondary SPI
+// for example does not support this Mode. You may need to use spi.Mode0
+// in this and similar cases.
+func New(p spi.Port, o *Opts, spiMode spi.Mode) (*Dev, error) {
+	c, err := p.Connect(20*physic.MegaHertz, spiMode, 8)
 	if err != nil {
 		return nil, err
 	}
--- a/apa102/apa102.go
+++ b/apa102/apa102.go
@@ -36,10 +36,11 @@ const NeutralTemp uint16 = 6500
 
 // DefaultOpts is the recommended default options.
 var DefaultOpts = Opts{
-	NumPixels:        150,   // 150 LEDs is a common strip length.
-	Intensity:        255,   // Full blinding power.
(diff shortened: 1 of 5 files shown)


maruel:
Hi! Thanks for the contribution! Sorry I dropped it in my emails.

Frankly, I think I'd prefer to just switch to Mode0 and not make it an option. What is the downside?

maruel:
Ignore the test failure, I need to redo the way I do forward-testing.

alexgth:
I'm not sure wether or not there might be some weird behaviour on the LED side of things due to the clock polarity being inverted. Whenever I accessed less LEDs than i connected the one LED after the last one was bright white, which might be because of the LEDs protocol (more likely for my understanding) or because of the SPI Mode. I unfortunately can't test it as I don't have access to the hardware anymore.

sahib:
@maruel: Are we sure that no other user of this library uses mode 3?
@alexgth: :wave:

maruel:
Frankly I'd simply use Mode1 and see if anyone complains on the next release. Users can always revert back to the previous release if it breaks them and we will add a bool flag at that point.

sahib:
@maruel: if users note that something is broken soon enough :sweat_smile: I guess the approach in this PR has the potential to save some tears. I was initially asking because I darkly recalled that we tried some LED strips initially on the first SPI port of a rpi3  (current approach uses second port), where mode 3 (old default) was working, but mode 1 not. That led to my guess that changing the default will break things for people.

maruel:
Sorry for the delay, lots of things happening here.

Thanks for the data point about some LEDs not working with mode3. I'd still prefer to use a bool defaulting to false so that a constructing a Opt{} by hand still works. Right now with this change constructing a Opt{} will result in spi.Mode0, which doesn't work. Please document the flag to explain the specific use case, and the fact that some LEDs may not work when the bool is set to true. 

I've finally fixed the github actions checks, so please git fetch and git rebase first. Thanks!

maruel:
Revisiting this PR: I'd really prefer to use a bool instead of a spi.Mode; this way the default bool value in Go (false) is the right one.

gsexton:
@maruel I've read through the chain of this. I see your point about constructing a device by hand results in Mode0 which breaks some things. The PR does have a default value that provides Mode3. 

I think having a boolean (UseMode0) is kind of a awkward.  How would you like to see this use a boolean?

maruel:
> @maruel I've read through the chain of this. I see your point about constructing a device by hand results in Mode0 which breaks some things. The PR does have a default value that provides Mode3.
>
> I think having a boolean (UseMode0) is kind of a awkward. How would you like to see this use a boolean?

3.5 years ago that was my preference from the perspective that otherwise new users that create an Options struct will be broken. My default stance is that working code should not be broken unless there's a good reason to. I don't feel this pass the bar, hence the bool.

PR #18, opened 2021-10-02
ds18b20: support for DS18S20

bboehmke:
This change add support for the [DS18S20](https://datasheets.maximintegrated.com/en/ds/DS18S20.pdf) to the ds18b20 package.
I also added some tests for the temperature parsing based on values from the datasheets of the DS18B20 and the DS18S20.

Because the implementation of these 2 sensors are nearly identically I simply extended the `ds18b20` package.


ds18b20/doc.go
```diff
@@ -21,6 +21,7 @@
 // Datasheets
 //
 // https://datasheets.maximintegrated.com/en/ds/DS18B20-PAR.pdf
+// https://datasheets.maximintegrated.com/en/ds/DS18S20.pdf
```
maruel, line 24: Technically it's better to leave an empty line so it shows up better at https://periph.io/x/devices/v3/ds18b20. I realize the other link should be changed to https. The two can be done in a follow up.

maruel (approved):
Thanks for the change, and thank for taking the time to add a unit test!

PR #19, opened 2021-10-04
ds18b20: fixed conversion time for DS18S20 + cleanup

bboehmke:
This is a followup of #18.
This ensures the right conversion time is used and the resolution setting is skipped for the DS18S20. (I totally missed that in #18)

Also addresses the remarks of #18 regarding the doc.go.


maruel:
Thanks!

PR #20, opened 2021-10-12
firmata: addition of support for firmata protocol

imle:
I wanted to open this early as it is a significant change and I would really like some input before I get much deeper into this implementation. There are parts of this missing and I have not implemented some of the features yet. This is a work in progress.


bmxx80/bmxx80.go
```diff
@@ -20,6 +20,9 @@ import (
 	"periph.io/x/conn/v3/spi"
 )
 
+// I2CAddr i2c default address.
```
maruel, line 23: Can you make a separate PR for this and the file below? They are unrelated to firmata.

maruel:
Apologies for the delay. Thanks I'll try to look at it ASAP.

bmxx80/bmxx80.go
imle: Didn't mean for that to make it in, will move to another PR.

firmata/analog_mapping.go
```diff
@@ -0,0 +1,43 @@
+package firmata
```
maruel, line 1: For every files: please add copyright.

firmata/analog_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"bytes"
+	"fmt"
+)
+
+type AnalogMappingResponse struct {
```
maruel, line 8: For every public symbols: Please document or make private.

firmata/analog_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+	AnalogPinToDigital []uint8
+	DigitalPinToAnalog map[uint8]uint8
+}
+
+func ParseAnalogMappingResponse(data []byte) AnalogMappingResponse {
+	var response = AnalogMappingResponse{
```
maruel, line 14: For most var uses in this PR: var is unnecessary.

firmata/analog_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+		AnalogPinToDigital: []uint8{},
```
maruel, line 15: This is unnecessary.

firmata/analog_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+		if data[i] != CapabilityResponsePinDelimiter {
+			response.DigitalPinToAnalog[uint8(i)] = uint8(len(response.AnalogPinToDigital))
+			response.AnalogPinToDigital = append(response.AnalogPinToDigital, uint8(i))
+		}
+	}
+
+	return response
+}
+
+func (a AnalogMappingResponse) String() string {
```
maruel, line 29: I recommend a pointer receiver otherwise it makes a copy.

firmata/analog_mapping.go
```diff
@@ -0,0 +1,43 @@
...
+	str := bytes.Buffer{}
+	for analogPin, digitalPin := range a.AnalogPinToDigital {
+		_, _ = fmt.Fprintf(&str, "A%d: %d\n", analogPin, digitalPin)
```
maruel, line 32: Generally speaking String() return value doesn't contain \n. Please use Sprintf() plus normal concatenation, given the expected number of pins it'll be faster in practice.

firmata/client.go
```diff
@@ -0,0 +1,679 @@
...
+	Close() error
+}
+
+type Client struct {
+	board                 io.ReadWriteCloser
+	responseChannels      map[SysExCmd][]chan []byte
+	sysExListenerChannels map[SysExCmd]chan []byte
+
+	i2cListeners map[uint8]chan I2CPacket
+	i2cMU        sync.Mutex
```
maruel, line 71: Put the mutex above what it protects, not below.

firmata/client.go
```diff
@@ -0,0 +1,679 @@
...
+	OpenOneWireBus(p uint8) (bus onewire.BusCloser, err error)
+	SetOneWireListener(uint8, chan []byte) (release func(), err error)
+
+	GetPinName(uint8) string
+	GetPinFunctions(uint8) []pin.Func
+
+	Close() error
+}
+
+type Client struct {
```
maruel, line 65: Can you either export the interface or the struct but not both? I generally prefer the struct unless there's a clear reason like to make a fake/mock easier to use in practice.

firmata/client.go
```diff
@@ -0,0 +1,679 @@
...
+
+	responseChannel := make(chan FirmwareReport, 1)
+	go func() {
+		responseChannel <- ParseFirmwareReport(<-firmChannel)
+	}()
+
+	go func() {
+		err := c.responseWatcher()
+		if err != nil {
+			panic(err)
```
maruel, line 155: Can you make this not be a panic?

firmata/util.go
```diff
@@ -0,0 +1,114 @@
+package firmata
```
maruel, line 1: name this file "conv.go", it's more meaningful than util.go.

pca9685/pins.go
```diff
@@ -96,7 +96,7 @@ func (p *pin) Read() gpio.Level {
 }
 
 func (p *pin) WaitForEdge(timeout time.Duration) bool {
-	return false
+	return gpio.INVALID.WaitForEdge(timeout)
```
maruel, line 99: Is that intended?

maruel (commented):
Thanks for the PR! That's a lot of functionality, that's really cool.

maruel:
What I would recommend is to start with only one subsystem, let's say i2c, and go from there. The PR is fairly large and will make the back and forth more difficult.

PR #22, opened 2021-10-22
mfrc522 - Fix 4-byte tag UIDs

### LINKED ISSUE #21: MFRC522 - ReadUID() returns 5 bytes for a tag with a 4-byte UID
**Describe the bug**
When reading the UID of a tag with a 4-byte UID `ReadUID()` returns 5 bytes.

**To Reproduce**
1. Call `ReadUID()` and scan an NFC tag with a 4-byte UID.
2. 5 bytes will be returned. Comparing to the tag's known UID, the fifth byte appears to be extraneous.

**Expected behavior**
4 bytes should be returned.

 - OS: Raspbian 10.9
 - Board: Gen 1 Raspberry Pi Model B
 - Device: MFRC522

**Additional context**
The problem appears to be here: https://github.com/periph/devices/blob/47f3fec45df8376070823f45f7624567267335a8/mfrc522/mfrc522.go#L404-414

Both `antiColl()` and `antColll2()` check the crc byte but then also return it. In the case of a 4-byte UID `selectCard()` then returns the whole slice including the CRC, oops.

I think in the case of a 7-byte NUID it will also be wrong, the code correctly strips byte 0 (`0x88`), then leaves the crc from `antiColl()`, then strips the crc from `antiColl2()`, meaning it will return 8-bytes but byte 4 will be extraneous.

brackendawson:
4-byte UIDs incorrectly included a 5th csum byte. 7-byte UIDs were unaffected.

fixes #21

### DIFF
--- a/mfrc522/mfrc522.go
+++ b/mfrc522/mfrc522.go
@@ -46,11 +46,11 @@ type config struct {
 
 type configF func(*config) *config
 
-// WithFixedUID sets the card reader to return correct 4-byte UIDs. Without
-// this setting 4-byte UIDs will be 5-bytes long with bytes 0 to 3 being the
-// correct UID and byte 4 being an XOR of bytes 0 to 3. 7-byte UIDs are correct
-// regardless of this configuration.
-func WithFixedUID() configF {
+// WithBogusUID sets the card reader to return incorrect 4-byte UIDs. In
+// version 3.6.12 and earlier this package ruturned 5-bytes for tags with a
+// 4-byte UID with bytes 0 to 3 being the correct UID and byte 4 being an XOR
+// of bytes 0 to 3. 7-byte UIDs are correct regardless of this configuration.
+func WithBogusUID() configF {
 	return func(c *config) *config {
 		c.bogusUID = false
 		return c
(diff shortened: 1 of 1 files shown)


mfrc522/mfrc522.go
```diff
@@ -40,10 +41,22 @@ type config struct {
 	defaultTimeout time.Duration
 	beforeCall     func()
 	afterCall      func()
+	bogusUID       bool
 }
 
 type configF func(*config) *config
 
+// WithFixedUID sets the card reader to return correct 4-byte UIDs. Without
```
maruel, line 49: After thinking more, I think it's better to switch the default and have the user explicitly opt in.

mfrc522/mfrc522.go
```diff
@@ -66,15 +79,20 @@ func WithSync() configF {
 func noop() {}
 
 // NewSPI creates and initializes the RFID card reader attached to SPI.
+// It is recommended to use the WithFixedUID configuration to return correct
```
maruel, line 82: Then you can skip this comment.

mfrc522/mfrc522.go
brackendawson: I'm happy with that, it also means we wouldn't need to update either periph/cmd or periph/website to include the option in the examples.

brackendawson:
Sorry about the force push, it turns out accidentally transposing two characters in my last commit message makes it massively inappropriate. 😳

maruel (approved):
lgtm albeit I'm a bit concerned that no unit test was touched. I'd be super happy if you can add one as a follow up but won't block on that.

maruel:
No worry about the force push, I squash anyway.

PR #24, opened 2021-11-10
mfrc522: The default should be not bogus

brackendawson:
Related to #21 I screwed up when we changed the default behaviour from bogus to not-bogus and overlooked the default value, it must be false because there is only an option to set it to true.


maruel:
Oops, I missed it. Thanks!

Issue #26, opened 2021-11-29
Display driver viewable via HTTP


hansmi:
**What kind of new feature are you looking for?**

In the interest of simplifying development on a host machine I implemented a [`display.Drawer`](https://pkg.go.dev/periph.io/x/periph/conn/display#Drawer) which is also an HTTP request handler. All client requests get an initial snapshot of the graphics buffer and are updated further on every change.

The protocol used is "MJPEG" (`multipart/x-mixed-replace` with images) which is often also used by IP cameras. Because of its better suitability for computer-drawn graphics the PNG image format is used by default, but JPEG can be selected as a default or using a URL parameter.

The only dependencies are the standard library and periph.io/display.

A snapshot of the source code is available at <https://github.com/hansmi/periph-webdisplay> with an example clock implementation at <https://github.com/hansmi/periph-webdisplay/tree/main/example/clock>. Go documentation is at <https://pkg.go.dev/github.com/hansmi/periph-webdisplay> (still a bit underpolished). Unit tests ensure that everything works.

I think such a driver would be useful to have in periph.io itself. I'm filing this feature request to see whether there's interest for opening a pull request and to discuss a few things before I finalize the documentation and API.

If a target device has a network interface (e.g. a Raspberry Pi), the proposed driver can also be used to access the display content remotely. I intend to write another driver combining multiple displays into one interface, distributing updates to all.

**Questions:**

I'm currently naming the driver `webdisplay`. Does anyone have a better name in mind?

Is `github.com/periph/devices` the most suitable repository?

**Do you plan to:**
- Contribute an initial driver: Yes
- Write unit tests: Yes
- Update https://github.com/periph/cmd to use the new functionality: The driver is not very useful on its own, but I could be convinced otherwise.

maruel:
Yes github.com/periph/devices for the generic implementation and github.com/periph/cmd for an executable sounds good. Thanks!

If I were to implement it personally, I would probably name it something like "VideoSink" or the "VideoServing" but I don't have a strong opinion.

PR #27, opened 2021-12-01
Add videosink display driver

hansmi:
The videosink package provides a display driver implementing an HTTP
request handler. Client requests get an initial snapshot of the graphics
buffer and are updated further on every change.

Provides the basic implementation for #26. Still to be done are example code and `periph/cmd` integration.

### DIFF
--- a/videosink/display.go
+++ b/videosink/display.go
@@ -1,0 +1,102 @@
+// Copyright 2021 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package videosink provides a display driver implementing an HTTP request
+// handler. Client requests get an initial snapshot of the graphics buffer and
+// are updated further on every change.
+//
+// The primary use case is the development of display outputs on a host
+// machine. Additionally devices with network connectivity can use this driver
+// to provide a copy of their local display via a web interface.
+//
+// The protocol used is "MJPEG" (https://en.wikipedia.org/wiki/Motion_JPEG)
+// which is often used by IP cameras. Because of its better suitability for
+// computer-drawn graphics the PNG image format is used by default. JPEG as
+// a format can be selected via Options.Format or using the "format" URL
+// parameter.
+package videosink
+
+import (
+	"image"
+	"image/color"
+	"image/draw"
+	"net/http"
+	"sync"
+
+	"periph.io/x/conn/v3/display"
+)
+
+// Options for videosink devices.
+type Options struct {
+	// Width and height of the image buffer.
+	Width, Height int
+
+	// Format specifies the image format to send to clients.
+	Format ImageFormat
+
+	// TODO: Add options for JPEG and PNG encoder settings
+}
+
+type Display struct {
+	defaultFormat ImageFormat
+
+	mu       sync.Mutex
+	buffer   *image.RGBA
+	clients  map[*client]struct{}
+	snapshot map[imageConfig][]byte
+}
+
+var _ display.Drawer = (*Display)(nil)
+var _ http.Handler = (*Display)(nil)
+
+// New creates a new videosink device instance.
+func New(opt *Options) *Display {
+	buffer := image.NewRGBA(image.Rect(0, 0, opt.Width, opt.Height))
+
+	// By default the alpha channel is set to full transparency. The following
+	// draw operation makes it opaque.
+	draw.Draw(buffer, buffer.Bounds(), image.Black, image.Point{}, draw.Src)
+
+	return &Display{
+		buffer:        buffer,
+		clients:       map[*client]struct{}{},
+		snapshot:      map[imageConfig][]byte{},
+		defaultFormat: opt.Format,
+	}
+}
+
+// String returns the name of the device.
+func (d *Display) String() string {
+	return "VideoSink"
+}
+
+// Halt implements conn.Resource and terminates all running client requests
+// asynchronously.
+func (d *Display) Halt() error {
+	d.mu.Lock()
+	d.terminateClientsLocked()
+	d.mu.Unlock()
+
+	return nil
+}
+
+// ColorModel implements display.Drawer.
+func (d *Display) ColorModel() color.Model {
+	return d.buffer.ColorModel()
+}
+
+// Bounds implements display.Drawer.
+func (d *Display) Bounds() image.Rectangle {
+	return d.buffer.Bounds()
+}
+
+// Draw implements display.Drawer.
+func (d *Display) Draw(dstRect image.Rectangle, src image.Image, srcPts image.Point) error {
+	d.mu.Lock()
+	draw.Draw(d.buffer, dstRect, src, srcPts, draw.Src)
+	d.bufferChangedLocked()
+	d.mu.Unlock()
+
+	return nil
+}
--- a/videosink/encoder.go
+++ b/videosink/encoder.go
@@ -1,0 +1,46 @@
+// Copyright 2021 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package videosink
+
+import (
+	"image/jpeg"
+	"image/png"
+	"sync"
+)
+
+var jpegOptions = jpeg.Options{
+	Quality: 95,
+}
+
+type pngEncoderBufferPool sync.Pool
+
+func (p *pngEncoderBufferPool) Get() *png.EncoderBuffer {
+	buf, _ := (*sync.Pool)(p).Get().(*png.EncoderBuffer)
+	return buf
+}
+
+func (p *pngEncoderBufferPool) Put(buf *png.EncoderBuffer) {
+	(*sync.Pool)(p).Put(buf)
+}
+
+type pngEncoderManager struct {
+	once sync.Once
+	pool pngEncoderBufferPool
+	enc  *png.Encoder
+}
+
+var pngEncoder pngEncoderManager
+
+// get returns a PNG encoder with a globally shared buffer pool.
+func (m *pngEncoderManager) get() *png.Encoder {
+	m.once.Do(func() {
+		m.enc = &png.Encoder{
+			CompressionLevel: png.BestSpeed,
+			BufferPool:       &m.pool,
+		}
+	})
+
+	return m.enc
+}
--- a/videosink/format.go
+++ b/videosink/format.go
@@ -1,0 +1,53 @@
+// Copyright 2021 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package videosink
+
+import "fmt"
+
+type ImageFormat int
+
+const (
+	PNG ImageFormat = iota
+	JPEG
+
+	// DefaultFormat is the format used when not set explicitly in options or
+	// as a URL parameter.
+	DefaultFormat = PNG
+)
+
+func (f ImageFormat) String() string {
+	switch f {
+	case PNG:
+		return "PNG"
+	case JPEG:
+		return "JPEG"
+	default:
+		return fmt.Sprint(int(f))
+	}
+}
+
+func (f ImageFormat) mimeType() string {
+	switch f {
+	case PNG:
+		return "image/png"
+	case JPEG:
+		return "image/jpeg"
+	}
+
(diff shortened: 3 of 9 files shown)


videosink/display.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2021 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// The videosink package provides a display driver implementing an HTTP request
```
maruel, line 5: // Package videosink provides ...

See https://go.dev/blog/godoc

videosink/display.go
```diff
@@ -0,0 +1,102 @@
...
+// Options for videosink devices.
+type Options struct {
+	// Width and height of the image buffer.
+	Width, Height int
+
+	// Format specifies the image format to send to clients.
+	Format ImageFormat
+}
+
+type Display struct {
```
maruel, line 39: I don't think Display is a good name, also document.

videosink/display.go
```diff
@@ -0,0 +1,102 @@
...
+	buffer   *image.RGBA
+	clients  map[*client]struct{}
+	snapshot map[imageConfig][]byte
+}
+
+var _ display.Drawer = (*Display)(nil)
+var _ http.Handler = (*Display)(nil)
+
+// New creates a new videosink device instance.
+func New(opt Options) *Display {
```
maruel, line 52: Take a pointer to Options

videosink/display.go
```diff
@@ -0,0 +1,102 @@
...
+	buffer := image.NewRGBA(image.Rect(0, 0, opt.Width, opt.Height))
+	draw.Draw(buffer, buffer.Bounds(), image.Black, image.Point{}, draw.Src)
```
maruel, line 54: It's unnecessary, it's the default.

videosink/encoder.go
```diff
@@ -0,0 +1,51 @@
...
+	"sync"
+)
+
+var jpegOptions = jpeg.Options{
+	Quality: 95,
+}
+
+type pngEncoderBufferPool sync.Pool
+
+var _ png.EncoderBufferPool = (*pngEncoderBufferPool)(nil)
```
maruel, line 19: I don't think it is useful.

videosink/encoder.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (p *pngEncoderBufferPool) Get() *png.EncoderBuffer {
+	if buf := (*sync.Pool)(p).Get(); buf != nil {
```
maruel, line 22: Can you confirm if this works?

```
buf, _ := (*sync.Pool)(p).Get().(*png.EncoderBuffer)
return buf
```

videosink/encoder.go
```diff
@@ -0,0 +1,51 @@
...
+	enc  *png.Encoder
+}
+
+var pngEncoder pngEncoderManager
+
+// get returns a PNG encoder with a globally shared buffer pool.
+func (m *pngEncoderManager) get() *png.Encoder {
+	m.once.Do(func() {
+		m.enc = &png.Encoder{
+			CompressionLevel: png.BestSpeed,
```
maruel, line 45: I would expect this to be configurable in Options eventually, I can imagine case where folks may want to have maximum compression.

videosink/format.go
```diff
@@ -0,0 +1,53 @@
...
+const (
+	PNG ImageFormat = iota
+	JPEG
+
+	// DefaultFormat is the format used when not set explicitly in options or
+	// as a URL parameter.
+	DefaultFormat = PNG
+)
+
+func (f ImageFormat) String() string {
```
maruel, line 20: I think I'd prefer to use stringer; golang.org/x/tools/cmd/stringer

videosink/handler.go
```diff
@@ -0,0 +1,179 @@
...
+
+	return cfg, nil
+}
+
+type client struct {
+	refresh   chan struct{}
+	terminate chan struct{}
+}
+
+func (d *Display) bufferChangedLocked() {
```
maruel, line 51: I would prefer to have all methods in the same file where the struct is defined. Otherwise one has to scan through multiple files to discover all the private methods (public ones can be found via godoc).

videosink/handler.go
```diff
@@ -0,0 +1,179 @@
...
+	for {
+		payload, err := d.grabSnapshot(cfg)
+		if err == nil {
+			err = pw.writeFrame(partHeaders, payload)
+			//lint:ignore SA6002 buffer is []byte and thus pointer-like
+			bufferPool.Put(payload)
+		}
+
+		if err != nil {
+			log.Printf("Request failed: %v", err)
```
maruel, line 160: I would prefer no log statement here.

videosink/handler.go
```diff
@@ -0,0 +1,179 @@
...
+	var err error
+
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	encoded, ok := d.snapshot[cfg]
+	if !ok {
+		encoded, err = d.encodeBufferLocked(cfg.format)
+		if err != nil {
+			return nil, err
```
maruel, line 97: The only case where it could fail if if there's an encoder configuration error or the display type is corrupted. I think you can just return nil here.

videosink/writer.go
```diff
@@ -0,0 +1,74 @@
...
+//
+// Go has a writer for MIME multipart messages in "mime/multipart".Writer. As
+// of Go 1.17 it's not suitable for writing a neverending stream of parts where
+// each must be flushed to the client with the part-ending boundary line.
+func (w *partWriter) writeFrame(header textproto.MIMEHeader, body []byte) error {
+	header.Set("Content-Length", strconv.FormatInt(int64(len(body)), 10))
+
+	var buf bytes.Buffer
+
+	if !w.started {
```
maruel, line 52: Why do you create a temporary buffer, to reduce the number of write calls to the response?

maruel (commented):
Thanks for the change! Nothing structural, just minutia questions.

.github/workflows/test.yml
```diff
@@ -216,7 +216,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        gover: ['1.13.15']
+        gover: ['1.14']
```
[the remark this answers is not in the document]
hansmi: It's described in the message for the first commit in this PR: <https://github.com/periph/devices/pull/27/commits/a5388a739159bd78c45bf5e928036941b390076e> (there are, as of this writing, two commits).

videosink/display.go
hansmi: Done.

videosink/display.go
hansmi: Do you have a better name in mind? `Drawer` as in my first round, or maybe `Sink`? `videosink.VideoSink` reads like [unnecessary stuttering](https://stackoverflow.com/a/45886055).

videosink/display.go
hansmi: Done.

videosink/display.go
hansmi: Not quite; the difference is in the alpha channel. `buffer.At(0, 0)` before that call to `draw.Draw` is `color.RGBA{R:0x0, G:0x0, B:0x0, A:0x0}`, after it's `color.RGBA{R:0x0, G:0x0, B:0x0, A:0xff}`. This doesn't matter for JPEG, but it does for PNG. Maybe we should leave this to the caller, however. What do you think?

videosink/encoder.go
hansmi: Removed the interface assertion.

videosink/encoder.go
hansmi: Yes, it does and I changed the code. During development I originally wrote `return (*sync.Pool)(p).Get().(*png.EncoderBuffer)` at first, but that triggers a panic at runtime: `interface conversion: interface {} is nil, not *png.EncoderBuffer`.

videosink/handler.go
hansmi: I agree with you and usually do the same. In this particular instance I elected to deviate and keep client-related code in `handler.go`. Well, mostly, because I just noticed that I added client-related code to `Display.Halt`.

Are you fine with the separation given the explanation or should I combine the code again?

videosink/handler.go
hansmi: What would you prefer to have instead? `fmt.Fprintf(os.Stderr, …)`, a configurable logger function/interface in the options, or silently ignoring?

videosink/handler.go
hansmi: I think even a panic here would be fine and still not hide the problem. `grabSnapshot` is called from the HTTP request handler for which Go's HTTP server handles panics. What do you think?

videosink/writer.go
hansmi: This function was heavily inspired by Go's own MIME multipart writer:

https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/mime/multipart/writer.go;l=97-130;drc=a6293995c55659f51e0662e7656f395633c99b5b

Yes, it does reduce write calls. Looking at it now we could also use the buffer pool here to avoid an allocation (though I'd first check if Go's escape analysis didn't eliminate that already).

Are you concerned about something in specific?

.github/workflows/test.yml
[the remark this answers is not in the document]
maruel: Let's do that as a separate CL, so we can update go.mod too.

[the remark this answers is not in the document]
maruel: By CL, I mean PR.

videosink/display.go
maruel: Technically, it's a stream of images, not really a video. images.Sink is a possibility.

My main concern is that it's really not a display. :) I agree we should try to not stutter.

videosink/display.go
maruel: oh good point!

videosink/handler.go
maruel: I don't feel the files are long enough to warrant splitting, but I don't feel strongly about that.

videosink/handler.go
maruel: Add a comment for now and silently ignore. We can add later. I prefer the code to not log by default.

videosink/handler.go
maruel: Generally not fond of panic but I think it's a fine choice here.

videosink/writer.go
maruel: No, just curious.

.github/workflows/test.yml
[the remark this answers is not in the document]
hansmi: Done in #29.

videosink/format.go
hansmi: I gave it a spin, only to realize that the `mimeType()` and `ImageFormatFromString` functions also need to be maintained when changing formats. As such I think stringer would add plenty of unnecessary complexity.

videosink/handler.go
hansmi: Done.

videosink/handler.go
hansmi: Done.

videosink/handler.go
hansmi: If you don't mind I'll leave it split it up then. That way the per-file focus is retained and the tests match up too.

videosink/encoder.go
hansmi: Added a TODO to that effect in the Options structure. I'll do it in a follow-up change, just like keep-alive and rate-limiting.

videosink/display.go
hansmi: You're right that it's currently a stream of images. Technically it would be possible to plug in a video encoder and stream the result of that to the HTTP client. I didn't do this because there's no video codec encoder in the standard library and images should be fine for development.

To be honest I'm not a fan of `Sink`. The _thing_ is also a virtual device, so the structure could be called `Device`. On the other hand it's an actual HTTP request handler masquerading as a display driver for periph.io, so `Handler` would also fit.

Which would you prefer?

Naming is hard.

maruel: I think I prefer Device to Display but it's a bit misleading. Given it's misleading, I'm fine with sticking to Display and moving on.

maruel:
I decided to merge, let's continue the discussion if we want to rename things a bit before next release. Even then it's not a big deal if we break it while there isn't users yet.

PR #28, opened 2021-12-01
waveshare2in13v2: Remove dependencies on fogleman/gg and golang/freetype

hansmi:
The waveshare2in13v2 example was evidently copied from the epd example,
but didn't comment out the code pulling in additional dependencies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,8 +7,6 @@ module periph.io/x/devices/v3
 go 1.13
 
 require (
-	github.com/fogleman/gg v1.3.0
-	github.com/golang/freetype v0.0.0-20170609003504-e2365dfdc4a0
 	github.com/maruel/ansi256 v1.0.2
 	github.com/mattn/go-colorable v0.1.11
 	golang.org/x/image v0.0.0-20210628002857-a66eb6448b8d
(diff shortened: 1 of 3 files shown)


maruel:
On one hand, I don't mind too much because it's in a test, thus it is not actually used anywhere. On the other hand, it's true that it means everyone fetches more than needed.

I reread https://go.dev/doc/go1.17#go-command and realized the package has to have a specific `go 1.17` to use the new [module graph pruning](https://go.dev/ref/mod#graph-pruning) which is a bit unfortunate.

hansmi:
Oh, I wasn't aware of pruned module graphs in 1.17. Thank you for pointing that out!

I think long-term that'll be the better path (with being provided by upstream) than the alternative of treating examples as their own packages and directories with a local `go.mod` and `go.sum`.

If I'm reading the [`go.mod` documentation for the `go` directive](https://go.dev/ref/mod#go-mod-file-go) correctly it would be perfectly fine to update to a 1.17 `go.mod`. Older versions would continue to download all dependencies while 1.17 and newer would prune the graph.

Of course this is a more generic topic than this particular PR.

PR #29, opened 2021-12-03
Bump Go version in go.mod and for quick tests to 1.14

hansmi:
Go 1.13 was released in September 2019, more than two years ago.
Updating to one later release should be fine.

The primary reason for doing this is the introduction of
"testing".T.Cleanup in Go 1.14 (https://pkg.go.dev/testing#T.Cleanup).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -216,7 +216,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        gover: ['1.13.15']
+        gover: ['1.14.15']
     runs-on: "${{matrix.os}}"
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -216,7 +216,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        gover: ['1.13.15']
+        gover: ['1.14']
```
maruel, line 219: Let's be explicit and use go1.14.15

hansmi: Done.

maruel:
Thanks!

PR #33, opened 2021-12-09
waveshare2in13v2: Deduplicate and simplify logic for sending data

hansmi:
Dev.Clear: Avoid writing one beyond the end of the line. Also build the
row data only once before sending it for each line.

Dev.Draw, Dev.DrawPartial: Deduplicate logic for iterating over pixels
and simplify it such that a whole row of bits is built before sending
them in one go.

Tested using a Waveshare e-Paper 2.13in V2 display.

### DIFF
--- a/waveshare2in13v2/waveshare213v2.go
+++ b/waveshare2in13v2/waveshare213v2.go
@@ -5,6 +5,7 @@
 package waveshare2in13v2
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


maruel:
Please fix the test failure then I can submit.

hansmi:
@maruel, sorry, missed that. Done.

PR #34, opened 2021-12-10
sgp30: add support for SGP30 air quality sensor

### LINKED ISSUE #32: Support SGP30 over I2C

- Hardware: [SGP30](https://shop.pimoroni.com/products/sgp30-air-quality-sensor-breakout)

I implemented in this [repository](https://github.com/0x46616c6b/airquality-exporter/blob/main/devices/sgp30/sgp30.go) the SGP30 already and want to ask if I can contribute this piece of code. It is fully functional but I did not implemented the CRC checks which are mentioned in the datasheet.

**Do you plan to:**
- Contribute an initial driver: Yes
- Write unit tests: No
- Update https://github.com/periph/cmd to use the new functionality: Yes

0x46616c6b:
- Fixes #32

@maruel feel free to add suggestions which commands should exported or if some documentation should be added.

### DIFF
--- a/sgp30/sgp30.go
+++ b/sgp30/sgp30.go
@@ -1,0 +1,192 @@
+// Copyright 2021 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package sgp30
+
+import (
+	"context"
+	"encoding/binary"
+	"errors"
+	"log"
+	"strconv"
+	"sync"
+	"time"
+
+	"periph.io/x/conn/v3"
+	"periph.io/x/conn/v3/i2c"
+)
+
+const (
+	initAirQuality       uint16 = 0x2003
+	measureAirQuality    uint16 = 0x2008
+	getIAQBaseline       uint16 = 0x2015
+	setIAQBaseline       uint16 = 0x201e
+	setHumidity          uint16 = 0x2061
+	measureTest          uint16 = 0x2032
+	getFeatureSetVersion uint16 = 0x202f
+	measureRawSignals    uint16 = 0x2050
+	getTVOCBaseline      uint16 = 0x20b3
+	setTVOCBaseline      uint16 = 0x2077
+
+	i2CAddress uint16 = 0x58
+)
+
+// commandDuration maps the defined maximum measurement duration from the sensor
+var commandDuration = map[uint16]time.Duration{
+	initAirQuality:       time.Millisecond * 10,
+	measureAirQuality:    time.Millisecond * 12,
+	getIAQBaseline:       time.Millisecond * 10,
+	setIAQBaseline:       time.Millisecond * 10,
+	setHumidity:          time.Millisecond * 10,
+	measureTest:          time.Millisecond * 220,
+	getFeatureSetVersion: time.Millisecond * 10,
+	measureRawSignals:    time.Millisecond * 25,
+	getTVOCBaseline:      time.Millisecond * 10,
+	setTVOCBaseline:      time.Millisecond * 10,
+}
+
+// commandResponseLength maps the defined response length including the CRC
+var commandResponseLength = map[uint16]int{
+	measureAirQuality:    6,
+	getIAQBaseline:       6,
+	measureTest:          3,
+	getFeatureSetVersion: 3,
+	measureRawSignals:    6,
+	getTVOCBaseline:      3,
+}
+
+// CO2 represents the current carbon dioxide value in ppm
+type CO2 uint16
+
+func (c CO2) String() string {
+	return strconv.Itoa(int(c)) + "ppm"
+}
+
+func (c *CO2) set(b []byte) {
+	*c = (CO2)(binary.BigEndian.Uint16(b))
+}
+
+// TVOC represents the current total volatile organic compounds value in ppb
+type TVOC uint16
+
+func (t TVOC) String() string {
+	return strconv.Itoa(int(t)) + "ppb"
+}
+
+func (t *TVOC) set(b []byte) {
+	*t = (TVOC)(binary.BigEndian.Uint16(b))
+}
+
+// Env represents measurements from an environmental sensor.
+type Env struct {
+	CO2  CO2
+	TVOC TVOC
+}
+
+// NewI2C returns an object that communicates over I2C to SGP30 environmental sensor.
+//
+// The address must be 0x58.
+func NewI2C(b i2c.Bus, ctx context.Context) (*Dev, error) {
+	d := &Dev{
+		d: &i2c.Dev{Bus: b, Addr: i2CAddress},
+		env: Env{
+			CO2:  400,
+			TVOC: 0,
+		},
+	}
+	if err := d.makeDev(ctx); err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
+// Dev is a handle to an initialized SGP30 device.
+type Dev struct {
+	d   conn.Conn
+	mu  sync.Mutex
+	env Env
+}
+
+// AirQuality return the value struct for the sensor
+func (d *Dev) AirQuality() Env {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	return d.env
+}
+
+func (d *Dev) makeDev(ctx context.Context) error {
+	// Sending  a "sgp30_iaq_init" command starts the air quality measurement
+	if err := d.initAirQuality(); err != nil {
+		return err
+	}
+
+	// After the "sgp30_iaq_init" command, a "sgp30_measure_iaq" command has to be sent in regular
+	// intervals of 1s to ensure proper operation of the dynamic baseline compensation algorithm.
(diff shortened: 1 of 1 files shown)


sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+	"log"
+	"strconv"
+	"time"
+
+	"periph.io/x/conn/v3"
+	"periph.io/x/conn/v3/i2c"
+)
+
+const (
+	InitAirQuality       uint16 = 0x2003
```
maruel, line 19: Please do not export these. I'd also vote for making it an explicit type.

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// CO2 represents the current carbon dioxide value in ppm
+type CO2 uint16
+
+func (c CO2) String() string {
+	return strconv.Itoa(int(c)) + "ppm"
+}
+
+func (c *CO2) Set(b []byte) {
```
maruel, line 64: This doesn't need to be exported.

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// TVOC represents the current total volatile organic compounds value in ppb
+type TVOC uint16
+
+func (t TVOC) String() string {
+	return strconv.Itoa(int(t)) + "ppb"
+}
+
+func (t *TVOC) Set(b []byte) {
```
maruel, line 75: Same

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+	return d, nil
+}
+
+// Dev is a handle to an initialized SGP30 device.
+type Dev struct {
+	d   conn.Conn
+	env Env
+}
+
+// EquivalentCO2 returns the latest equivalent CO2 value
```
maruel, line 113: Why not expose Env instead of having methods? This could read as the last read value. But this has an inherent race condition, so I recommend against.

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// TotalVOC returns the latest total VOC value
+func (d *Dev) TotalVOC() TVOC {
+	return d.env.TVOC
+}
+
+func (d *Dev) makeDev() error {
+	// Sending  a "sgp30_iaq_init" command starts the air quality measurement
+	err := d.initAirQuality()
```
maruel, line 125: Inline in next line.

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	// After the "sgp30_iaq_init" command, a "sgp30_measure_iaq" command has to be sent in regular
+	// intervals of 1s to ensure proper operation of the dynamic baseline compensation algorithm.
+	if err := d.measure(); err != nil {
+		log.Print(err)
+	}
+
+	ticker := time.NewTicker(1 * time.Second)
```
maruel, line 136: Can you make it a separate function that accepts a context.Context? I'm thinking about migrating everything to this pattern for v4.

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+				log.Print(err)
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (d *Dev) initAirQuality() error {
+	var err error
```
maruel, line 150:
```
err := d.writeCommand(InitAirQuality)
if err == nil {
  time.Sleep(time.Second * 20)
}
return err
```

or

```
if err := d.writeCommand(InitAirQuality); err != nil {
  return err
}
time.Sleep(time.Second * 20)
return nil
```

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+	if err = d.writeCommand(InitAirQuality); err == nil {
+		time.Sleep(time.Second * 20)
+	}
+
+	return err
+}
+
+func (d *Dev) measure() error {
+	buf := make([]byte, commandResponseLength[MeasureAirQuality])
+	err := d.readCommand(MeasureAirQuality, buf)
```
maruel, line 160: Inline

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return err
+	}
+	d.env.CO2.Set(buf[0:2])
```
maruel, line 164: Or return a Env here.

sgp30/sgp30.go
```diff
@@ -0,0 +1,194 @@
...
+	if len(b) != commandResponseLength[cmd] {
+		return errors.New("response length mismatch")
+	}
+
+	regAddr := []byte{byte(cmd >> 8), byte(cmd & 0xFF)}
+	if err := d.d.Tx(regAddr, nil); err != nil {
+		return err
+	}
+	time.Sleep(commandDuration[cmd])
+	if err := d.d.Tx(nil, b); err != nil {
```
maruel, line 180: return d.d.Tx(nil, b)

maruel (commented):
Thanks!

0x46616c6b:
I addressed most of your suggestions. The change for `Env` I don't really get or don't have a better idea. What do you think?

sgp30/sgp30.go
```diff
@@ -0,0 +1,193 @@
...
+// Env represents measurements from an environmental sensor.
+type Env struct {
+	CO2  CO2
+	TVOC TVOC
+}
+
+// NewI2C returns an object that communicates over I2C to SGP30 environmental sensor.
+//
+// The address must be 0x58.
+func NewI2C(b i2c.Bus, addr uint16, ctx context.Context) (*Dev, error) {
```
maruel, line 89: Why have addr as a parameter if it must be one value?

sgp30/sgp30.go
maruel: Ignore.

sgp30/sgp30.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Dev is a handle to an initialized SGP30 device.
+type Dev struct {
+	d   conn.Conn
+	env Env
+}
+
+// EquivalentCO2 returns the latest equivalent CO2 value
+func (d *Dev) EquivalentCO2() CO2 {
+	return d.env.CO2
```
maruel, line 116: FTR, this is a race condition. I would prefer a single method that returns an Env.

sgp30/sgp30.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return err
+}
+
+func (d *Dev) measure() error {
+	buf := make([]byte, commandResponseLength[measureAirQuality])
+	if err := d.readCommand(measureAirQuality, buf); err != nil {
+		return err
+	}
+	d.env.CO2.set(buf[0:2])
```
maruel, line 166: Use a sync.Mutex.

sgp30/sgp30.go
```diff
@@ -0,0 +1,193 @@
...
+func (d *Dev) EquivalentCO2() CO2 {
+	return d.env.CO2
+}
+
+// TotalVOC returns the latest total VOC value
+func (d *Dev) TotalVOC() TVOC {
+	return d.env.TVOC
+}
+
+func (d *Dev) makeDev(ctx context.Context) error {
```
maruel, line 124: I do plan to change all drivers to use this form. I think I'd prefer it to exported as ReadLoop(ctx) and be blocking. Then it's up to the caller to use a go routine instead of the driver.

sgp30/sgp30.go
0x46616c6b: Should I implement the `Env` (or with another name) and Units in `periph.io/x/conn/v3/physic`? Seems I implemented it in another way it was implemented for example for bmxx80 sensors (which I orientated at).

maruel: No, it's fine to keep a local Env struct as you did.

0x46616c6b: I removed the methods and return the struct instead.

sgp30/sgp30.go
0x46616c6b: Done

sgp30/sgp30.go
0x46616c6b: Do you also plan to add an interface for the devices? Should I implement this now or will you (or should I) adjust this later?

maruel: It can be done later, it's fine for now.

sgp30/sgp30.go
```diff
@@ -0,0 +1,189 @@
...
+// Dev is a handle to an initialized SGP30 device.
+type Dev struct {
+	d   conn.Conn
+	mu  sync.Mutex
+	env Env
+}
+
+// AirQuality return the value struct for the sensor
+func (d *Dev) AirQuality() Env {
+	return d.env
```
maruel, line 113: Please acquire the mutex here, otherwise it's a race condition.

0x46616c6b: Done

maruel (approved):
Thanks!

PR #38, opened 2021-12-18
waveshare213v2: Add support for partial drawing

hansmi:
A series of changes bringing support for partial updates to the `waveshare213v2` driver. Unittests are implemented to verify that the calculation of the various offsets is correct.

At the moment the partial updates are only useful when aligned to 8-pixel boundaries on the horizontal axis. A future series will bring flicker-free updates with pixel precision using the second image buffer as a mask (it's already working in an experimental implementation).


maruel:
Thanks!

PR #43, opened 2021-12-27
videosink: Implement rate-limiting, keep-alive

### LINKED ISSUE #26: Display driver viewable via HTTP

hansmi:
Resolve the remaining two TODOs by adding an upper limit on the frame
rate (defaults to 15 per second) and sending an image every once in
a while even when there were no changes (defaults to once a minute).

Fixes #26.

### DIFF
--- a/videosink/display.go
+++ b/videosink/display.go
@@ -25,11 +25,14 @@ import (
 	"image/png"
 	"net/http"
 	"sync"
+	"time"
 
 	"periph.io/x/conn/v3/display"
 )
 
 const defaultJPEGQuality = 95
+const defaultKeepAliveInterval = time.Minute
+const defaultMinFrameInterval = time.Second / 15
 
 // Options for videosink devices.
 type Options struct {
@@ -48,6 +51,16 @@ type Options struct {
 		// encoder. Defaults to png.DefaultCompression.
 		CompressionLevel png.CompressionLevel
 	}
+
+	// MinFrameInterval is the amount of time which needs to pass before
+	// sending a new image, thus implementing rate-limiting. Defaults to 15
+	// frames per second.
+	MinFrameInterval time.Duration
+
+	// KeepAliveInterval is the amount of time after which to send a new
+	// image regardless of whether any changes have been made. Defaults to
+	// once per minute.
+	KeepAliveInterval time.Duration
 }
 
 // Display is a virtual device receiving drawing operations and sending
@@ -56,6 +69,8 @@ type Display struct {
 	defaultFormat       ImageFormat
 	jpegOptions         jpeg.Options
 	pngCompressionLevel png.CompressionLevel
+	keepAliveInterval   time.Duration
+	minFrameInterval    time.Duration
 
 	mu       sync.Mutex
 	buffer   *image.RGBA
@@ -77,6 +92,8 @@ func New(opt *Options) *Display {
 	d := &Display{
 		jpegOptions:         opt.JPEG,
 		pngCompressionLevel: opt.PNG.CompressionLevel,
+		keepAliveInterval:   opt.KeepAliveInterval,
+		minFrameInterval:    opt.MinFrameInterval,
 
 		buffer:        buffer,
 		clients:       map[*client]struct{}{},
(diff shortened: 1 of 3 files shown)


videosink/display.go
```diff
@@ -25,11 +25,14 @@ import (
...
 	"net/http"
 	"sync"
+	"time"
 
 	"periph.io/x/conn/v3/display"
 )
 
 const defaultJPEGQuality = 95
+const defaultKeepAliveInterval = time.Minute
+const defaultMinFrameInterval = time.Second / 15
```
maruel, line 35: You may want `(time.Second + time.Second/30) / 15`

videosink/handler.go
```diff
@@ -23,6 +25,17 @@ var bufferPool = sync.Pool{
 	},
 }
 
+// stopAndDrain ensures that the given timer is stopped and has no pending
+// event.
+func stopAndDrain(timer *time.Timer) {
+	if !(timer == nil || timer.Stop()) {
```
maruel, line 31: I feel this is more readable: `if timer != nil && !timer.Stop() {`

videosink/handler.go
```diff
@@ -44,8 +57,63 @@ func (d *Display) configFromQuery(values url.Values) (imageConfig, error) {
...
+
+// Wait until the next frame should be sent, either because the keep-alive
+// interval has passed or a change has been made to the buffer and the rate
+// limit isn't violated.
+func (c *client) waitNext(ctx context.Context) bool {
+	earliestFrameAt := c.mostRecent.Add(c.disp.minFrameInterval)
+	latestFrameAt := c.mostRecent.Add(c.disp.keepAliveInterval)
+
+	var rateLimitTimer *time.Timer
+	var keepAliveTimer = time.NewTimer(time.Until(latestFrameAt))
```
maruel, line 84: `keepAliveTimer := time.NewTimer(time.Until(latestFrameAt))`

videosink/handler.go
```diff
@@ -44,8 +57,63 @@ func (d *Display) configFromQuery(values url.Values) (imageConfig, error) {
...
+		select {
+		case <-rateLimit:
+			return true
+
+		case <-c.refresh:
+			if remaining := time.Until(earliestFrameAt); remaining <= 0 {
+				return true
+			} else if rateLimitTimer == nil {
+				rateLimitTimer = time.NewTimer(remaining)
+				defer stopAndDrain(rateLimitTimer)
```
maruel, line 104: You could make rateLimit a function local variable (instead of being inside the for loop) if you would set its value here. I think it'd be more readable.

videosink/display.go
hansmi: What would be the practical benefit? Per https://go.dev/play/p/aadQgiosT-z your proposal and my code differ by ~2.2ms—not relevant.

videosink/handler.go
hansmi: Changed even though my personal preference is to avoid chains with multiple negations.

videosink/handler.go
hansmi: Done.

videosink/handler.go
hansmi: Good idea; done.

PR #45, opened 2022-01-18
Minor fixes for waveshare2in13v2, fix typos in image1bit

hansmi:
* Add tests for `waveshare2in13v2.New()`
* Swap the order of width and height in `waveshare2in13v2.Dev.String()`
* Simplify memory configuration


maruel:
Thanks!

PR #46, opened 2022-01-21
waveshare2in13v2: Support rotation of image

hansmi:
A new option named `Origin` controls which of the four corners of the display should be used as the (0,0) position for the image. The included unittests may seem excessive, but they are the result of many off-by-one errors and other mistakes I made while developing this series.

### DIFF
--- a/waveshare2in13v2/drawing.go
+++ b/waveshare2in13v2/drawing.go
@@ -52,21 +52,21 @@ type drawOpts struct {
 
 type drawSpec struct {
 	// Destination on display in pixels, normalized to fit into actual size.
-	DstRect image.Rectangle
+	dstRect image.Rectangle
 
 	// Area to send to device; horizontally in bytes (thus aligned to
 	// 8 pixels), vertically in pixels.
-	MemRect image.Rectangle
+	memRect image.Rectangle
 }
 
 func (o *drawOpts) spec() drawSpec {
 	s := drawSpec{
-		DstRect: image.Rectangle{Max: o.devSize}.Intersect(o.dstRect),
+		dstRect: image.Rectangle{Max: o.devSize}.Intersect(o.dstRect),
 	}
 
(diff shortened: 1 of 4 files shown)


waveshare2in13v2/drawing.go
```diff
@@ -44,54 +44,150 @@ func setMemoryArea(ctrl controller, area image.Rectangle) {
...
+	// Amount by which buffer contents must be moved to align with the physical
+	// top-left corner of the display.
+	//
+	// TODO: The offset shifts the buffer contents to be aligned such that the
+	// translated position of the physical, on-display (0,0) location is at
+	// a multiple of 8 on the equivalent to the physical X axis. With a bit of
+	// additional work transfers for the TopRight and BottomLeft origins should
+	// not require per-pixel processing by exploiting image1bit.VerticalLSB's
+	// underlying pixel storage format.
+	BufferDstOffset image.Point
```
maruel, line 64: If you don't intent to export the struct, it'd be good to not export its fields.

waveshare2in13v2/waveshare213v2.go
```diff
@@ -77,19 +77,30 @@ type Dev struct {
...
 	busy gpio.PinIn
 
+	bounds image.Rectangle
 	buffer *image1bit.VerticalLSB
 	mode   PartialUpdate
 
 	opts *Opts
 }
 
+type Corner uint8
```
maruel, line 87: Please document.

waveshare2in13v2/drawing.go
hansmi: Done.

waveshare2in13v2/waveshare213v2.go
hansmi: Done.

maruel (approved):

PR #47, opened 2022-04-15
Support for sh1106 display.

nlacasse:
Taken from:
https://github.com/sztanpet/sh1106

Which itself is mostly taken from `ssd1306/ssd1306.go`.

I only updated the dependencies to point to the new locations.

### DIFF
--- a/sh1106/sh1106.go
+++ b/sh1106/sh1106.go
@@ -1,0 +1,484 @@
+// Copied almost verbatim from https://github.com/google/periph/tree/master/devices/ssd1306
+// The sh1106 differs from the ssd1306 only in writing to it's memory.
+// We now also properly use the startCol and endCol values from calculateSubset.
+//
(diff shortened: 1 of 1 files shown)


sh1106/sh1106.go
```diff
@@ -0,0 +1,484 @@
+// Copied almost verbatim from https://github.com/google/periph/tree/master/devices/ssd1306
```
maruel, line 1: Shouldn't it be https://github.com/periph/devices/tree/main/ssd1306 ?

maruel (commented):
Apologies for the delay, I was travelling.
I'll check later why go1.18 is used in the CI; I know that gosec is not yet compatible.

I'll have to checkout and diff with the other driver to see the difference. It'd be nice to not duplicate code but it's not a big deal.

There's no doc nor tests.

gsexton:
Implemented in https://github.com/periph/devices/pull/104

PR #49, opened 2022-09-23
epd: added support for waveshare 2in13 rev2.1

mhazley:
Added a new set of Opts for the Waveshare epd 2in13 rev2.1 - needed a different LUT.


maruel:
Thanks!

PR #50, opened 2022-09-28
tca95xx: Add support for tca95xx i/o extenders.

schmidtw:
1. :thumbsup:
2. n/a
3. :thumbsup:

### Things of note
- I am able to test the TCA9534 and TCA9535 chips using a RPI4.  I included chips that were nearly identical as to reduce the likelihood of introducing a bug.
- I added a conn.Conn interface and implemented that for the extender to allow access to the entire port as a byte.
- This code is based on the `mcp23xxx` code pretty heavily.


schmidtw:
I noticed the copyright dates were wrong.  Opening a new pr.

PR #51, opened 2022-09-28
Add support for tca95xx i/o extenders.

schmidtw:
1. :thumbsup:
2. n/a
3. :thumbsup:
### Things of note
- I am able to test the TCA9534 and TCA9535 chips using a RPI4.  I included chips that were nearly identical as to reduce the likelihood of introducing a bug.
- I added a conn.Conn interface and implemented that for the extender to allow access to the entire port as a byte.
- This code is based on the `mcp23xxx` code pretty heavily.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -38,4 +38,5 @@ Matias Insaurralde <matias@insaurral.de>
 Seán C McCord <ulexus@gmail.com> <scm@cycoresys.com>
 Stephan Sperber <sperberstephan@googlemail.com>
 Thorsten von Eicken <tve@voneicken.com>
+Weston Schmidt <weston_schmidt@alumni.purdue.edu>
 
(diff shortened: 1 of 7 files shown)


maruel:
Hi! Thanks for the contribution! It's really appreciated.
Would you mind fixing the staticcheck warning? It helps code readability overall.

schmidtw:
Can I add https://github.com/stretchr/testify to help me clean up a bunch of cruft in the unit tests?

maruel:
While I understand the value for those who are familiar with this kind of library, a package like https://pkg.go.dev/github.com/stretchr/testify/assert is not what I call a great design. It's so large that it cannot fit in my head, which means that I have to look up every single function when I use it, or try to guess what it does based on the name. Feel free to create a one-off local function in the test file to reduce copy pasting.

schmidtw:
Is there anything else I need to do before you can merge?

maruel:
Thanks a lot for the PR. Apologies for the delay, work was busy this week.

PR #52, opened 2022-09-30
tlv493d: Auto-reset TLV493D on error

bezineb5:
When an error occurs, it is necessary to reset the sensor as per the documentation. This will send a recovery command followed by the standard start procedure.

### DIFF
--- a/tlv493d/tlv493d.go
+++ b/tlv493d/tlv493d.go
@@ -405,14 +405,16 @@ func (d *Dev) ReadContinuous(frequency physic.Frequency, precision Precision) (<
 				if err != nil {
 					// Try resetting the sensor to recover from errors
 					fmt.Println("Error reading from TLV493D sensor:", err)
-					if err := d.initialize(true); err != nil {
-						continue
+					if err := d.initialize(true); err == nil {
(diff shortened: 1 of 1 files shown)


tlv493d/tlv493d.go
```diff
@@ -403,7 +403,15 @@ func (d *Dev) ReadContinuous(frequency physic.Frequency, precision Precision) (<
 			case <-t.C:
 				value, err := d.Read(precision)
 				if err != nil {
-					// In continuous mode, we'll ignore errors silently.
+					// Try resetting the sensor to recover from errors
+					fmt.Println("Error reading from TLV493D sensor:", err)
```
maruel, line 407: I'm uncomfortable with leaving fmt.Println() in the code. If you really feel it needs to be logged you can send it to log.Printf() but I'd prefix it with the device name so that if there's multiple devices the user can differentiate.

bezineb5: Oups, indeed, sorry for that. I kept logs in case of reset as it might indicate an electrical/I2C bus issue.

maruel:
Thanks!

PR #53, opened 2022-10-02
tlv493d: improved reset logging

bezineb5:
Sorry, this is a follow-up for #52 , with proper log formatting.

### DIFF
--- a/tlv493d/tlv493d.go
+++ b/tlv493d/tlv493d.go
@@ -405,14 +405,16 @@ func (d *Dev) ReadContinuous(frequency physic.Frequency, precision Precision) (<
 				if err != nil {
 					// Try resetting the sensor to recover from errors
 					fmt.Println("Error reading from TLV493D sensor:", err)
-					if err := d.initialize(true); err != nil {
-						continue
+					if err := d.initialize(true); err == nil {
+						if err := d.SetMode(newMode); err != nil {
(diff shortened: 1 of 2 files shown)


tlv493d/tlv493d.go
```diff
@@ -407,9 +407,9 @@ func (d *Dev) ReadContinuous(frequency physic.Frequency, precision Precision) (<
 					// Try resetting the sensor to recover from errors
 					if err := d.initialize(true); err == nil {
 						if err := d.SetMode(newMode); err != nil {
-							log.Println("Unable to reset TLV493D mode:", err)
+							log.Printf("%s: unable to reset tlv493d mode: %v", d, err)
```
maruel, line 410: One problem is that the String() doesn't include the port so it wouldn't help diagnosing as much as if it included the connection.

bezineb5: Makes sense, I added the i2c address in the String(). However, it seems that there's no common pattern in all the logging and name methods, maybe a shared template could be enforced at least for the new drivers? Or maybe I just overlooked it :D

maruel: No driver really log anything at the moment. It was intentional but I agree it can be annoying to diagnose issues. The system would benefit from being more opiniated here.

maruel:
Thanks!

PR #54, opened 2022-11-14
ina219: Interpret signed output register values

eldondevat:
The current and shunt voltage register values are signed, twos complement format. In my tests interpreting them directly in the way implemented here gave output consistent with other ina219 libraries.

### DIFF
--- a/ina219/ina219.go
+++ b/ina219/ina219.go
@@ -5,6 +5,7 @@
 package ina219
 
 import (
+	"unsafe"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


ina219/ina219.go
```diff
@@ -110,7 +111,7 @@ func (d *Dev) Sense() (PowerMonitor, error) {
 		return PowerMonitor{}, errReadShunt
 	}
 	// Least significant bit is 10µV.
-	pm.Shunt = physic.ElectricPotential(shunt) * 10 * physic.MicroVolt
+	pm.Shunt = physic.ElectricPotential(*((*int16)(unsafe.Pointer(&shunt)))) * 10 * physic.MicroVolt
```
maruel, line 114: Why not int16(shunt) ? No need for unsafe. See https://go.dev/play/p/uUhKx4hdbeR

eldondevat: I typically write go at an abstraction level a bit higher, and to be honest, the 0xFFFF result surprised me. Seems perfectly reasonable, to use that instead though.

PR #57, opened 2023-03-26
lirc requries linebreak \n for SEND_ONCE

### LINKED ISSUE #56: LIRC requires line break for SEND_ONCE
**Describe the bug**
Dunno since when but lirc requires a linebreak for SEND_ONCE or It will throw `bad send packet` error
You can refer the source code https://github.com/aldebaran/lirc/blob/master/daemons/lircd.c#L774

**To Reproduce**
1. Run program
```go
package main

import (
  "periph.io/x/host/v3"
)

func main() {
  irConn, err := lirc.New()
  err = irConn.Emit("RemoteConfiguredInLIRC", ir.Key("CommandConfiguredInLIRC"))
}
```
2. Run it.
3. See error
Check syslog and find 
`bad send packet`
Or no command was logged as triggered in syslog

**Expected behavior**
The SEND_ONCE should be triggered and there should have no error like "bad send packet"

 - OS: Ubuntu 20.04
 - Board Raspberry Pi 4

**Additional context**
Confirmed that change the line lirc.go:67 from 
`_, err := fmt.Fprintf(c.w, "SEND_ONCE %s %s", remote, key)`
to
`_, err := fmt.Fprintf(c.w, "SEND_ONCE %s %s\n", remote, key)`
will work

emulatorchen:
Fixes periph.io/x/devices/v3/devices/lirc/lirc.go:67

- Fixes #56

P.S. There's already another issue when the command is sent successfully but the parse of lirc.go cannot recognize it the the SEND_ONCE trigger a raw config. I will fill another issue once I have the solution.

### DIFF
--- a/lirc/lirc.go
+++ b/lirc/lirc.go
@@ -64,7 +64,7 @@ func (c *Conn) Close() error {
 // Emit implements ir.IR.
 func (c *Conn) Emit(remote string, key ir.Key) error {
 	// http://www.lirc.org/html/lircd.html#lbAH
-	_, err := fmt.Fprintf(c.w, "SEND_ONCE %s %s", remote, key)
+	_, err := fmt.Fprintf(c.w, "SEND_ONCE %s %s\n", remote, key)
 	return err
 }
 


maruel:
Thanks!

PR #66, opened 2023-12-28
adxl345: Driver, initial commit for review.

benoit-pereira-da-silva:
if this PR fixes periph.io/x/devices/v3/cap1xxx, prefix the PR title with "cap1xxx:".

- Added basic support for the adxl345 accelerometer.

3. Once integrated, send a PR to https://github.com/periph/cmd to leverage the
new functionality (if relevant).

### DIFF
--- a/adxl345/adxl345.go
+++ b/adxl345/adxl345.go
@@ -152,19 +152,20 @@ func (d *Dev) TurnOff() error {
 // This is a simple synchronous implementation.
 func (d *Dev) Update() Acceleration {
 	return Acceleration{
-		X: d.readAndCombine(DataX0, DataX1),
-		Y: d.readAndCombine(DataY0, DataY1),
-		Z: d.readAndCombine(DataZ0, DataZ1),
+		X: d.ReadAndCombine(DataX0, DataX1),
+		Y: d.ReadAndCombine(DataY0, DataY1),
+		Z: d.ReadAndCombine(DataZ0, DataZ1),
 	}
 }
 
-// readAndCombine combines two registers to form a 16-bit value.
-// Example:
-// `DATAX0` is the address of the lower byte (LSB, least significant byte) of the X-axis data
-// `DATAX1` is the address of the upper byte (MSB, most significant byte) of the X-axis data
+// ReadAndCombine combines two registers to form a 16-bit value.
+// The ADXL345 uses two 8-bit registers to store the output data for each axis.
+// X := d.ReadAndCombine(DataX0, DataX1) where:
+// `DataX0` is the address of the lower byte (LSB, least significant byte)
+// `DataX1` is the address of the upper byte (MSB, most significant byte)
 // The ADXL345 combines both registers to deliver 16-bit output for each acceleration axis.
 // A similar approach is used for the Y and Z axes. This technique provides higher precision in the measurements.
-func (d *Dev) readAndCombine(reg1, reg2 byte) int16 {
+func (d *Dev) ReadAndCombine(reg1, reg2 byte) int16 {
 	low, _ := d.Read(reg1)
 	high, _ := d.Read(reg2)
 	return int16(uint16(high)<<8) | int16(low)
--- a/adxl345/adxl345.go
+++ b/adxl345/adxl345.go
@@ -7,10 +7,15 @@ import (
 	"periph.io/x/conn/v3/spi"
 )
 
+type Sensitivity byte
+
 const (
 	DeviceID = 0x00 // Device ID, expected to be 0xE5 when using ADXL345
 
-	// 0x01 to 0x1C  are reserved for future use
+	S2G  Sensitivity = 0x00 // Sensitivity at 2g
+	S4G  Sensitivity = 0x01 // Sensitivity at 4g
+	S8G  Sensitivity = 0x02 // Sensitivity at 8g
+	S16G Sensitivity = 0x03 // Sensitivity at 16g
 
 	ThreshTap        = 0x1D // Tap threshold
 	OfsX             = 0x1E // X-axis offset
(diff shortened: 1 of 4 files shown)


adxl345/adxl345.go
```diff
@@ -0,0 +1,230 @@
...
+	"encoding/binary"
+	"fmt"
+	"periph.io/x/conn/v3/physic"
+	"periph.io/x/conn/v3/spi"
+)
+
+type Sensitivity byte
+
+const (
+	DeviceID = 0x00 // Device ID, expected to be 0xE5 when using ADXL345
```
maruel, line 13: Please do not export these unless they are meant to be used by the user.

adxl345/adxl345.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// FIFO control
+	FifoCtl    = 0x38 // FIFO control
+	FifoStatus = 0x39 // FIFO status
+
+)
+
+// DefaultSpiFrequency is the default SPI frequency used to communicate with the device.
+var (
+	SpiFrequency = physic.KiloHertz * 50
```
maruel, line 62: 50kHz is extremely slow for SPI?

Please inline these values, no need to have variables.

adxl345/adxl345.go
```diff
@@ -0,0 +1,230 @@
...
+	// Convert the spi.Port into a spi.Conn so it can be used for communication.
+	c, err := p.Connect(SpiFrequency, SpiMode, SpiBits)
+	if err != nil {
+		return nil, err
+	}
+	d := &Dev{
+		name: "ADXL345",
+		s:    c,
+	}
+	if o.TurnOnOnStart {
```
maruel, line 106: I don't think it's necessary, it's okay to have users call it manually, or always do it by default.

adxl345/adxl345.go
```diff
@@ -0,0 +1,230 @@
...
+// The ADXL345 combines both registers to deliver 16-bit output for each acceleration axis.
+// A similar approach is used for the Y and Z axes. This technique provides higher precision in the measurements.
+func (d *Dev) ReadAndCombine(reg1, reg2 byte) int16 {
+	low, _ := d.Read(reg1)
+	high, _ := d.Read(reg2)
+	return int16(uint16(high)<<8) | int16(low)
+}
+
+// Read reads a 16-bit value from the specified register address.
+func (d *Dev) Read(regAddress byte) (int16, error) {
```
maruel, line 175: I'd recommend to use https://pkg.go.dev/periph.io/x/conn/v3@v3.7.0/mmr#Dev16.ReadUint16 instead and convert back to a Int16.

adxl345/example/example.go
```diff
@@ -0,0 +1,53 @@
...
+	// Use spireg SPI port registry to find the first available SPI bus.
+	p, err := spireg.Open("")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	defer p.Close()
+
+	o := adxl345.DefaultOpts
+	d, err := adxl345.New(p, &o)
```
maruel, line 30: d, err := adxl345.New(p, &adxl345.DefaultOpts)

maruel (commented):
Thanks for the PR!

adxl345/adxl345.go
benoit-pereira-da-silva: It is a Register it is normal to keep it public for anyone that wants to read it.

adxl345/adxl345.go
benoit-pereira-da-silva: You are absolutely right. I ve fixed 2Gz , ADXL2345 supports up to 5gz.

adxl345/adxl345.go
benoit-pereira-da-silva: Removed.

adxl345/example/example.go
benoit-pereira-da-silva: Done

adxl345/adxl345.go
benoit-pereira-da-silva: @maruel Sorry, I m quite new to "periph" and i do not understand

```go
func (d *Dev) Read(regAddress byte) (int16, error) {
	tx := []byte{regAddress | 0x80, 0x00}
	rx := make([]byte, len(tx))
	err := d.s.Tx(tx, rx)
	if err != nil {
		return 0, err
	}
	return int16(binary.LittleEndian.Uint16(rx)), nil
}``` 
this conversion is equivalent to `return int16(uint16(rx[0]) | uint16(rx[1])<<8)`  i feel that using Dev16.ReadUint16 won't be more efficient.

Comment (maruel) [reply adxl345/adxl345.go:181]: It's okay. The point is to reduce the amount of code as much as possible but it's not a big deal.

**Comment by maruel** [comment]
Please fix the CI error.

**Comment by benoit-pereira-da-silva** [comment]
> Please fix the CI error
I m sorry but i can't figure out what is that error 

 ```Initializing: exit status 128:
fatal: cannot mkdir /home/cuong/go/src/github.com/benoit-pereira-da-silva/devices-0: No space left on device
```

maruel:
Oh, I meant https://github.com/periph/devices/actions/runs/7348677329/job/20007433704?pr=66

adxl345/adxl345.go
```diff
@@ -0,0 +1,252 @@
...
+
+type Opts struct {
+	ExpectedDeviceID byte        // Expected device ID used to verify that the device is an ADXL345.
+	Sensitivity      Sensitivity // Sensitivity of the device (2G, 4G, 8G, 16G)
+}
+
+// Dev is a driver for the ADXL345 accelerometer
+// It uses the SPI interface to communicate with the device.
+type Dev struct {
+	c    mmr.Dev16
```
maruel, line 93: Remove it if you don't plant to use it.

adxl345/adxl345.go
```diff
@@ -0,0 +1,252 @@
...
+func (d *Dev) String() string {
+	return fmt.Sprintf("ADXL345{Sensitivity:%s}", d.sensitivity)
+}
+
+// New creates a new ADXL345 Dev or returns an error.
+// The bus and chip parameters define the SPI bus and chip select to use.
+// The SPI s is configured.
+// The device is turned on.
+// The device is verified to be an ADXL345.
+func New(p spi.Port, o *Opts) (*Dev, error) {
```
maruel, line 110: I'd recommend to name it NewSPI right away so if I2C is added later, it's a non-breaking change.

maruel (approved):
Two minor things then it's good to go.

adxl345/adxl345.go
benoit-pereira-da-silva: Done ...  Will try to be a better citizen next time I will use mmr when providing I2C support. This version as been well tested on real hardwares via SPI only, i don't want to provide I2C without verifying it.

adxl345/adxl345.go
benoit-pereira-da-silva: Done

PR #67, opened 2023-12-29
adxl345:  I²C support

benoit-pereira-da-silva:
if this PR fixes periph.io/x/devices/v3/cap1xxx, prefix the PR title with "cap1xxx:".
adxl345: support both SPI and I²C.
+ Refactored a little bit to make it more consistant with other devices implementation.

3. Once integrated, send a PR to https://github.com/periph/cmd to leverage the
new functionality (if relevant).


benoit-pereira-da-silva:
@maruel using mmr.Dev16 was not convenient. The adxl345 returns positive and negative int16 values. So i decided to keep the conversion the infamous :) conversion  `int16(binary.LittleEndian.Uint16(rx))`  this implementation should work for most of the adxlXXX line so i ve enabled the ability to use another device by passing explicitly its Opts.ExpectedDeviceID.

maruel:
Thanks for the change! Can you rebase it onto origin/main first?

adxl345/example/examples.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2023 The Periph Authors. All rights reserved.
```
maruel, line 1: I just realized you had put the example in a separate directory. This is not customary in Go. Any specific reason for this?

adxl345/example/examples.go
```diff
@@ -0,0 +1,96 @@
...
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package example
```
maruel, line 5: You can use `package adxl345_test` and it will show up great in the documentation.

adxl345/example/examples.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+)
+
+// I2C uses an adxl345 device connected by I²C.
+// it reads the acceleration values every 30ms for 30 seconds.
+// You can i use `i2dctools`  to find the I²C bus number
+// e.g : sudo apt-get install i2c-tools
+//
+//	sudo i2cdetect -y 1
+func I2C(addr uint16) {
```
maruel, line 23: You want it to be ExampleNewI2C() for it to show up at the right location in the documentation. See https://go.dev/blog/examples for more information.

adxl345/example/examples.go
benoit-pereira-da-silva: No valuable reason :)

adxl345/example/examples.go
benoit-pereira-da-silva: Done

adxl345/example/examples.go
benoit-pereira-da-silva: Thank you for pointing out "testable examples"  in fact My IDE was pointing out that the signature was not acceptable in the test package (see below) . In fact that why i created an "example" package.  I've been following your recommandation in my last commit. 
Screenshot 2024-01-02 à 16 24 33

adxl345/example_test.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2023 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package adxl345
```
maruel, line 5: Optional: if you use `package adxl345_test`, the example in the documentation becomes copy-pastable as-is.

maruel: I'll resolve to commit, this can be done in a follow up and is optional.

benoit-pereira-da-silva: I've been pushing the change.

PR #68, opened 2024-01-14
Add AHT20 sensor

SoulKa:
Implemented the AHT20 environmental sensor using the `SenseEnv` interface. Implementation is based on this [Datasheet](https://files.seeedstudio.com/wiki/Grove-AHT20_I2C_Industrial_Grade_Temperature_and_Humidity_Sensor/AHT20-datasheet-2020-4-16.pdf). Tested on a RaspberryPi with the AHT20 sensor.

**Features:**
- Pure Go, no dependencies
- `Sense()` reads a single measurement
- `SenseContinuous()` reads measurements with the given interval into a channel (or as fast as possible if it is too small)
- `Halt()` stops a continuous measurement
- Options:
  - A configurable polling interval if the measurement is not finished after 80ms. Defaults to 10ms
  - A configurable measurement read timeout after the initial delay of 80ms. Default is 150ms.
  - Optionally CRC8 data validation of measurements. Default is enabled

### DIFF
--- a/aht20/aht20.go
+++ b/aht20/aht20.go
@@ -208,7 +208,7 @@ func calculateCRC8(data []byte) uint8 {
 		crc ^= b
 		for i := 0; i < 8; i++ {
 			if crc&0x80 != 0 {
-				crc = (crc << 1) ^ crc8Polynomial // 0x07 is the polynomial
+				crc = (crc << 1) ^ crc8Polynomial
 			} else {
 				crc <<= 1
 			}
--- a/aht20/aht20.go
+++ b/aht20/aht20.go
@@ -102,10 +102,15 @@ func (d *Dev) Sense(e *physic.Env) error {
 		// read measurement
 		if err := d.d.Tx(nil, data); err != nil {
 			return err
-		} else if d.opts.ValidateData && calculateCRC8(data[0:6]) != data[6] {
-			return &DataCorruptionError{}
 		}
 
+		// validate data
+		dataCrc := calculateCRC8(data[0:6])
+		if d.opts.ValidateData && dataCrc != data[6] {
+			return &DataCorruptionError{Received: data[6], Calculated: dataCrc}
+		}
+
+		// check if measurement is ready
 		if data[0]&bitInitialized == 0 {
 			return &NotInitializedError{}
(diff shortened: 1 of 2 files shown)


aht20/aht20.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// NewI2C returns an object that communicates over I²C to AHT20 environmental sensor. The sensor
+// will be calibrated and initialized if it is not already. The Opts can be nil.
+func NewI2C(b i2c.Bus, opts *Opts) (*Dev, error) {
+	if opts == nil {
+		opts = &DefaultOpts
+	}
+	if opts.MeasurementWaitInterval <= 0 {
+		opts.MeasurementWaitInterval = 10 * time.Millisecond
```
maruel, line 69: I recommend not modifying Opts here, it can be surprising to users. You can modify after you made a copy at line 72.

maruel (commented):
Thanks for the new driver! Please resolve the github actions failure.

aht20/aht20.go
SoulKa: Good point. I changed it accordingly

SoulKa:
Thanks for the fast reaction on this PR! I resolved your issue and the pipeline issues. Somehow, `errors.Join()` is not known during build. Is it not built with Go 1.20? At least what's written in the go.mod.

Should I add some tests using the bus mock as in bmxx80?

maruel:
I try to keep compatibility with older version of Go since it's what is installed via apt-get install and users could be confused.

aht20/aht20.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// DefaultOpts holds the default configuration options for the device.
+var DefaultOpts = Opts{
+	MeasurementReadTimeout:  150 * time.Millisecond,
+	MeasurementWaitInterval: 10 * time.Millisecond,
+	ValidateData:            true,
+}
+
+// NewI2C returns an object that communicates over I²C to AHT20 environmental sensor. The sensor
```
maruel, line 61: It's okay to call it `New` since it's I²C only. The NewI2C naming is generally used when there's also a SPI version.

maruel (approved):
It's better if you can add tests but I won't require it.

PR #70, opened 2024-01-16
Add AHT20 tests

SoulKa:
- Add tests for all exported methods of the AHT20 device (except `SenseContinuous()`)
- Fix `IsInitialized()` always returning `false` due to data byte not being set


maruel:
Thanks! If you want one optional thing that is sweet is to add an example that can be copy pasted by using `package aht20_test`. See other example files for an example and the corresponding generated doc.

PR #72, opened 2024-01-16
Add doc file for AHT20

SoulKa:
Add `doc.go` for package `aht20` which I forgot in my last PR

### DIFF
--- a/aht20/doc.go
+++ b/aht20/doc.go
@@ -3,9 +3,15 @@
 // that can be found in the LICENSE file.
 
 // Package aht20 controls an AHT20 device over I²C.
+//
+// # More details
+//
 // The sensor is a temperature and humidity sensor with a typical accuracy of ±2% RH and ±0.3°C.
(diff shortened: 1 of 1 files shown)


aht20/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2024 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// Package aht20 controls an AHT20 device over I²C.
+// The sensor is a temperature and humidity sensor with a typical accuracy of ±2% RH and ±0.3°C.
```
maruel, line 6: if you want maximum polish, insert an empty line after the first line to create a paragraph
.

Here's how it can look like with titles; https://pkg.go.dev/periph.io/x/devices/v3@v3.7.1/apa102 and compare to the source file: https://github.com/periph/devices/blob/v3.7.1/apa102/doc.go.

SoulKa: changes applied, please check

PR #74, opened 2024-09-04
periph.io/x/devices/v3/bmxx80: add support for filters without using SenseContinuous

BGodding:
I cannot use the `SenseContinuous` option with my hardware as there are additional bus contention issues to deal with, but would like to use the built in filtering on the BME sensors. 

In order to prevent current implementations from changing their behavior I added a new initialization option to set the standby time. When the `Filter` and `Standby` options are set in the device options the sensor will initialize to the `normal` mode instead of `sleep`. Additionally when calling `Sense` mode will not be set to `forced` if the `Filter` and `Standby` options are set as this would prevent the continuous sampling needing for the filtration.

### DIFF
--- a/bmxx80/bmx280.go
+++ b/bmxx80/bmx280.go
@@ -75,7 +75,7 @@ const (
 // measurements are done.
 type standby uint8
 
-// Possible standby values, these determines the refresh rate.
+// Possible standby values, these determine the refresh rate.
 const (
 	s500us   standby = 0
 	s10msBME standby = 6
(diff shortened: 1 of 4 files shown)


maruel:
Dang gosec got new checks. Let me fix that first, shouldn't take too long.

maruel:
I committed https://github.com/periph/devices/commit/67ec1da96441ed538cd9faca0b01ff04b326d970 to silence the linter.

maruel:
Oh I forgot to mention: thanks for the contribution!

PR #76, opened 2024-10-06
tmp102: Initial add of tmp102 driver.

gsexton:
This is an implementation of a driver for the Texas Instruments TMP102 sensor. It's compatible with TMP112, and TMP75 as well.


maruel:
Thanks!

PR #77, opened 2024-10-20
ssd1306: Refactor display options

czechbol:
I hit an issue where I needed to horizontally mirror the image and I could not figure out how to do it with the current state of the implementation.

### DIFF
--- a/ssd1306/ssd1306_test.go
+++ b/ssd1306/ssd1306_test.go
@@ -29,7 +29,7 @@ func TestNewI2C_fail(t *testing.T) {
 	if d, err := NewI2C(&bus, &Opts{W: 64}); d != nil || err == nil {
 		t.Fatal(d, err)
 	}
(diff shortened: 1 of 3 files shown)


ssd1306/ssd1306_test.go
```diff
@@ -491,7 +491,7 @@ func TestSPI_3wire(t *testing.T) {
 func TestSPI_4wire_String(t *testing.T) {
 	port := spitest.Playback{
 		Playback: conntest.Playback{
-			Ops: []conntest.IO{{W: getInitCmd(&Opts{W: 128, H: 64, Rotated: false})}},
+			Ops: []conntest.IO{{W: getInitCmd(&Opts{W: 128, H: 64, MirrorVertical: false, MirrorHorizontal: false})}},
```
maruel, line 494: FYI, no need to add the properties set to false since it's the default. It's okay either way.

maruel (approved):
Thanks!

ssd1306/ssd1306_test.go
maruel: Resolving to merge.

PR #78, opened 2024-10-25
tic: Add I²C support for Tic Stepper Motor Controllers

z-riley:
This PR adds functionality for interfacing with [Pololu Tic Stepper Motor Controllers](https://www.pololu.com/category/212/tic-stepper-motor-controllers) over I²C.

The `tic` package is based on [github.com/pololu/tic-arduino](https://github.com/pololu/tic-arduino), with adoption of Go-style error handling, [periph/conn](https://pkg.go.dev/periph.io/x/periph/conn), and other usability improvements.

### DIFF
--- a/tic/registers.go
+++ b/tic/registers.go
@@ -1,0 +1,83 @@
+// Copyright 2024 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+package tic
+
+import (
+	"encoding/binary"
+)
+
+// getVar8 reads an 8 bit value from the Tic at a given register offset.
+func (d *Dev) getVar8(offset Offset) (uint8, error) {
+	const length = 1
+	buffer, err := d.getSegment(cmdGetVariable, offset, length)
+	if err != nil {
+		return 0, err
+	}
+
+	return buffer[0], nil
+}
+
+// getVar16 reads a 16 bit value from the Tic at a given register offset.
+func (d *Dev) getVar16(offset Offset) (uint16, error) {
+	const length = 2
+	buffer, err := d.getSegment(cmdGetVariable, offset, length)
+	if err != nil {
+		return 0, err
+	}
+
+	return binary.LittleEndian.Uint16(buffer), nil
+}
+
+// getVar32 reads a 32 bit value from the Tic at a given register offset.
+func (d *Dev) getVar32(offset Offset) (uint32, error) {
+	const length = 4
+	buffer, err := d.getSegment(cmdGetVariable, offset, length)
+	if err != nil {
+		return 0, err
+	}
+
+	return binary.LittleEndian.Uint32(buffer), nil
+}
+
+// commandQuick sends a command without additional data.
+func (d *Dev) commandQuick(cmd command) error {
+	writeBuf := []byte{uint8(cmd)}
+	err := d.c.Tx(writeBuf, nil)
+	return err
+}
+
+// commandW7 sends a command with a 7 bit value. The MSB of val is ignored.
+func (d *Dev) commandW7(cmd command, val uint8) error {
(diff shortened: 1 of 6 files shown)


tic/registers.go
```diff
@@ -0,0 +1,83 @@
...
+// that can be found in the LICENSE file.
+
+package tic
+
+import (
+	"encoding/binary"
+)
+
+// getVar8 reads an 8 bit value from the Tic at a given register offset.
+func (d *Dev) getVar8(offset Offset) (uint8, error) {
```
maruel, line 12: FYI, you could have used https://periph.io/x/conn/v3/mmr#Dev8 which takes care of all of this.

tic/registers.go
```diff
@@ -0,0 +1,83 @@
...
+// commandW7 sends a command with a 7 bit value. The MSB of val is ignored.
+func (d *Dev) commandW7(cmd command, val uint8) error {
+	writeBuf := []byte{byte(cmd), val & 0x7F}
+	err := d.c.Tx(writeBuf, nil)
+	return err
+}
+
+// commandW32 sends a command with a 32 bit value.
+func (d *Dev) commandW32(cmd command, val uint32) error {
+	writeBuf := make([]byte, 5)
```
maruel, line 60: FYI, this works:
```
writeBuf := [5]byte{byte(cmd)}
```

The only drawback is that line 64 would have to be replaced with:

```
err := d.c.Tx(writeBuf[:], nil)
```

tic/tic.go
```diff
@@ -0,0 +1,1330 @@
...
+func (d *Dev) SetMaxDecel(decel uint32) error {
+	return d.commandW32(cmdSetDecelMax, decel)
+}
+
+// StepMode describes how many microsteps add up to one fulls step.
+type StepMode uint8
+
+const (
+	StepModeFull StepMode = 0
+	StepModeHalf StepMode = 1
```
maruel, line 363: This looks strange that StepModeHalf == StepModeMicrostep2 ? I don't see these two values (including StepModeFull) being used anywhere.

tic/tic.go
```diff
@@ -0,0 +1,1330 @@
...
+// EEPROM, see the "Settings reference" section of the Tic user's guide.
+func (d *Dev) GetSetting(offset Offset, length uint) ([]uint8, error) {
+	if length > 15 {
+		return nil, errors.New("maximum length exceeded")
+	}
+
+	return d.getSegment(cmdGetSetting, offset, length)
+}
+
+// Offset represents where settings are stored in the Tic's EEPROM memory. See
```
maruel, line 1253: Unless mistaken, I didn't see the type "Offset" being used in any of the APIs. As such, I'd prefer to not export it as this will confuse the user. That's the only change I'd prefer to see before merging.

maruel (commented):
Thanks, this is excellent! The only blocking thing is that I believe Offset and its constants don't need to be exposed since they are not usable by the user.

tic/registers.go
z-riley: I wasn't aware of that package - very useful! However in this case, two bytes are sent to the device (`cmdGetVariable` + `offset`), so I'm struggling to see I'd use `mmr` to do this.

tic/registers.go
z-riley: [577d57f](https://github.com/periph/devices/pull/78/commits/577d57f1b0243ac545e2eaf81ffd6589f6503697)

tic/tic.go
z-riley: You're right - fixed in b0dd501.

tic/registers.go
maruel: Personally I wouldn't have changed the others, since the number of elements desired in the array matched the number of elements provided. But it's okay.

maruel:
Thanks!

PR #79, opened 2024-10-26
Max7219 LED Driver

gsexton:
This commit adds a driver for the Maxim 7219 LED Driver.


gsexton:
@maruel Is there something broken? It's showing a fail on codecov/project.

maruel:
Can you rebase on main? I just merged https://github.com/periph/devices/pull/80 which should fix the regression because I just reenabled codecov but the project coverage was too low.

max7219/example_test.go
```diff
@@ -0,0 +1,82 @@
...
+	var matrix = false
+	matrixUnits := 1
+	var err error
+	_, err = host.Init()
+	if err != nil {
+		log.Fatal(err)
+	}
+	if _, err = driverreg.Init(); err != nil {
+		log.Fatal(err)
+	}
```
maruel, line 30: I'd recommend:

```
if _, err := host.Init(); err != nil {
  log.Fatal(err)
}
```

driverreg.Init() is redundant. err gets defined at line 32 so no need to define it explicitly.

max7219/example_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"periph.io/x/conn/v3/driver/driverreg"
+	"periph.io/x/conn/v3/spi/spireg"
+	"periph.io/x/devices/v3/max7219"
+	"periph.io/x/host/v3"
+)
+
+// basic test program. To do a numeric display, set matrix to false and
+// matrixUnits to 1.
+func Example() {
+	var matrix = false
```
maruel, line 21:
```
matrix := false
```

max7219/max7219.go
```diff
@@ -0,0 +1,458 @@
...
+		w[ix*2+1] = data
+	}
+	return d.conn.Tx(w, nil)
+}
+
+// NewSPI creates a new Max7219 using the specified spi.Port. units is the number
+// of Max7219 chips daisy-chained together. numDigits is the number of digits
+// displayed.
+func NewSPI(p spi.Port, units, numDigits int) (*Dev, error) {
+	if numDigits <= 0 {
```
maruel, line 126: `|| numDigits/units > 8`

but the units <= 0 check would have to be done first.

max7219/max7219.go
```diff
@@ -0,0 +1,458 @@
...
+		return nil, errors.New("invalid value for number of digits")
```
maruel, line 127: I'd prefer the errors to be prefixed with "max7219: "

max7219/max7219.go
```diff
@@ -0,0 +1,458 @@
...
+	// DecodeNone is RAW mode, or not decoded. For each byte, bits that are
+	// one are turned on in the matrix, and bits that are 0 turn off the
+	// led at that row/column.
+	DecodeNone DecodeMode = 0
+)
+
+// Type for a Maxim MAX7219/MAX7221 device.
+type Dev struct {
+	conn spi.Conn
+	mu   sync.Mutex
```
maruel, line 55: I don't believe this device should be thread safe, I'd personally skip on the locking and would expect the caller to ensure locking. That's what most devices do.

max7219/max7219.go
```diff
@@ -0,0 +1,458 @@
...
+	units int
+	// The number of digits (or the NxN matrix size) in this display.
+	digits byte
+	// charset is the character set for matrix display use. The first index
+	// is the code point (e.g. 0x20=32=space. The second index is the 8 byte
+	// raster values for each line in the matrix.
+	glyphs [][]byte
+}
+
+// emptyBytes creates a slice of empty bytes (digit values or byte values)
```
maruel, line 68: If you stop trying to make the Dev thread safe, then this can be a property of Dev that you initialize once.

max7219/max7219.go
```diff
@@ -0,0 +1,458 @@
...
+		return d.WriteCascadedUnits(writeData)
+	}
+	return nil
+}
+
+// WriteInt provide a convenience method that displays the specified integer
+// value on the display. It will work for either matrixes (with a glyph set)
+// or numeric displays.
+func (d *Dev) WriteInt(value int) error {
+	var digits int
```
maruel, line 399:
```
digits := d.units
if d.decode != DecodeNone {
  digits *= int(d.digits)
}
```

maruel (approved):
Thanks! A few minors comments but nothing really blocking.

max7219/max7219.go
```diff
@@ -0,0 +1,440 @@
+// Copyright 2024 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+//
```
maruel, line 4: Oops, you need to not put the `//` here otherwise the copyright will appear in the package documentation.

max7219/max7219.go
```diff
@@ -0,0 +1,440 @@
...
+		{_REGISTER_DISPLAY_TEST, 0x0},
+		{_REGISTER_SHUTDOWN, 0x00},
+		{_REGISTER_INTENSITY, 0x08},
+		{_REGISTER_SCAN_LIMIT, d.digits - 1},
+		{_REGISTER_SHUTDOWN, 0x01}}
+
+	for _, cmd := range initCommands {
+		err := d.sendCommand(cmd[0], cmd[1])
+		if err != nil {
+			log.Println(err)
```
maruel, line 92: In a follow up, it'd be nicer to return the error here, and down below too.

max7219/max7219.go
```diff
@@ -0,0 +1,440 @@
...
+		return nil, errors.New("max7219: invalid value for number of digits")
+	}
+
+	// It works in Mode0, Mode2 and Mode3.
+	c, err := p.Connect(10*physic.MegaHertz, spi.Mode0, 8)
+	if err != nil {
+		return nil, fmt.Errorf("max7219: %v", err)
+	}
+	d := &Dev{conn: c, digits: byte(numDigits), units: units, glyphs: nil}
+	d.init()
```
maruel, line 135: then check for the return value and fail accordingly.

max7219/max7219.go
```diff
@@ -0,0 +1,440 @@
...
+		return nil, errors.New("max7219: invalid value for number of cascaded units")
+	}
+	if numDigits <= 0 || numDigits > 8 {
+		return nil, errors.New("max7219: invalid value for number of digits")
+	}
+
+	// It works in Mode0, Mode2 and Mode3.
+	c, err := p.Connect(10*physic.MegaHertz, spi.Mode0, 8)
+	if err != nil {
+		return nil, fmt.Errorf("max7219: %v", err)
```
maruel, line 132: In a follow up; you want `fmt.Errorf("max7219: %w", err)`. The %w means that the error will be properly wrapped. See https://go.dev/blog/go1.13-errors#wrapping-errors-with-w

max7219/max7219.go
```diff
@@ -0,0 +1,440 @@
...
+// SetIntensity controls the brightness of the display. The allowed range for
+// intensity is from 0-15. Keep in mind that the brighter display, the more
+// current drawn.
+func (d *Dev) SetIntensity(intensity byte) error {
+	return d.sendCommand(_REGISTER_INTENSITY, intensity&0x0f)
+}
+
+// TestDisplay turns on the 7219 display mode which set all segments (or LEDs) on,
+// and  the intensity to maximum. If you're using multiple units, you should be
+// aware  of the current draw, and limit how long you leave this on.
```
maruel, line 266: double spaces twice.

maruel (approved):
I noticed a few other items but they can be done in a follow up.

max7219/max7219.go
maruel: Resolving so I can merge.

max7219/max7219.go
maruel: Resolving so I can merge.

max7219/max7219.go
maruel: Resolving so I can merge.

max7219/max7219.go
maruel: Resolving so I can merge.

max7219/max7219.go
maruel: Resolving so I can merge.

maruel:
Thanks!

PR #81, opened 2024-11-08
scd4x: SCD4x CO2 Sensor - Initial Add

gsexton:
@maruel I'm mostly liking they way I did the unit tests. I get live device, and playback testing without lots of extra code.

What are your thoughts? Do you have any ideas on making it more elegant?

### DIFF
--- a/scd4x/scd4x_test.go
+++ b/scd4x/scd4x_test.go
@@ -1,0 +1,439 @@
+// Copyright 2024 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+//
+// Unit tests for the package. Note that this supports running on a live
+// sensor, or using playback mode to simulate a live device.
+//
+// To use a live device, define the environment variable SCD4X and run go test.
+
+package scd4x
+
+import (
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"periph.io/x/conn/v3/i2c"
+	"periph.io/x/conn/v3/i2c/i2creg"
+	"periph.io/x/conn/v3/i2c/i2ctest"
+	"periph.io/x/conn/v3/physic"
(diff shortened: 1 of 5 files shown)


maruel:
Thanks! Will look later.

scd4x/scd4x_test.go
```diff
@@ -0,0 +1,439 @@
...
+	{Addr: SensorAddress, W: []uint8{0x20, 0x2f}, R: []uint8{0x4, 0x41, 0xe}},
+	{Addr: SensorAddress, W: []uint8{0x23, 0x22}, R: []uint8{0x6, 0x44, 0x22}},
+	{Addr: SensorAddress, W: []uint8{0x23, 0x18}, R: []uint8{0x5, 0xda, 0x29}},
+	{Addr: SensorAddress, W: []uint8{0x36, 0x46}}}
+
+var basicStartup = []i2ctest.IO{
+	{Addr: SensorAddress, W: []uint8{0x36, 0xf6}},
+	{Addr: SensorAddress, W: []uint8{0x21, 0xb1}}}
+
+func init() {
```
maruel, line 97: Another way would be to use TestMain as described at https://pkg.go.dev/testing#hdr-Main.

```
func TestMain(m *testing.M) {
  flag.BoolVar(&liveDevice, "live", os.Getenv("SCD4X")!="", "Run unit test on a live device (a smoke test). Set automatically if env var SCD4X is set")
  m.Run()
}
```

maruel (approved):

maruel:
> @maruel I'm mostly liking they way I did the unit tests. I get live device, and playback testing without lots of extra code.
> 
> What are your thoughts? Do you have any ideas on making it more elegant?

Generally I've put these into cmd smoketest but I don't mind much, this is fine.

maruel:
None of the comments are blocking so I'll merge right away, you can do a follow up if you want but it's not required.

PR #83, opened 2024-12-07
Add TI HDC302x Sensor Support

gsexton:


maruel (approved):
Wow really cool! Thanks!

Dang the check failure in https://github.com/periph/devices/actions/runs/12216048416/job/34078745883?pr=83 is real but it's totally unrelated to your change. It's a data race in driver scd4x. Filed https://github.com/periph/devices/issues/84

gsexton:
> Wow really cool! Thanks!
> 
> Dang the check failure in https://github.com/periph/devices/actions/runs/12216048416/job/34078745883?pr=83 is real but it's totally unrelated to your change. It's a data race in driver scd4x. Filed #84

We crossed paths. The same race was happening in this branch. Also, there was a general race condition in the logic of the test that I found while doing hdc302x, so I fixed it in scd4x.

PR #86, opened 2024-12-20
ep0099: optionally read device state on new

freman:
Add new structure (Opts) and 2 new methods (NewWithOpts, ReadStates) to provide the ability to load the state from the device instead of simply resetting the device on start.

Promoted Reset() to be a public function so it can be called to reset the state if required.


maruel (approved):
Thanks!

PR #87, opened 2025-01-10
ina219: restructure so voltage can be returned even if power/current invalid

### LINKED ISSUE #73: ina219: Overflow bit indicates a math overflow error
**Describe the bug**
In [`ina219.go`](https://github.com/periph/devices/blob/main/ina219/ina219.go#L115-L122) there's a check for the overflow flag respresented in the voltage register. This error reads `bus voltage register overflow`. This is a misnomer.

This register is documented in the [datasheet](https://www.ti.com/lit/gpn/ina219) as follows:
<details>

</details>

```
Math Overflow Flag

The Math Overflow Flag (OVF) is set when the Power or Current calculations are out of range. It indicates that
current and power data may be meaningless.
```

This flag indicates not an overflow in the voltage register, but an overflow in the power and current registers, which are read out later in the code.

This is unfortunate, because the data in the voltage register is completely valid in such a state. The problem would be a misconfiguration in the calibration register, causing the internal calculations from shunt voltage to current and power to fail. The sad thing about this, is that a lot of precision is lost in this conversion, as the registers the chip is working with are very small (16 bits). I'd like to retrieve the Voltage value, and do the calculations myself in the code instead of on the chip.

**Expected behavior**
I would like the library to clarify the error message, and ideally provide a way to retrieve the measured full-scale voltage and shunt voltage regardless of this calculation overflow.

Personally I also think this library would be more useful calculating the current/power itself instead of relying on the chip, but that's out of the scope of this specific request.

**Additional context**
Device datasheet: https://www.ti.com/lit/gpn/ina219

gsexton:
Fixes #73

I would have wrapped the errors returned since I know that's what you prefer, but the tests would have been pretty broken and I didn't see a straight-forward repair.

### DIFF
--- a/ina219/ina219.go
+++ b/ina219/ina219.go
@@ -99,44 +99,48 @@ const (
 )
 
 // Sense reads the power values from the ina219 sensor.
-func (d *Dev) Sense() (PowerMonitor, error) {
+func (d *Dev) Sense() (pm PowerMonitor, err error) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
-	var pm PowerMonitor
-
 	shunt, err := d.m.ReadUint16(shuntVoltageRegister)
 	if err != nil {
-		return PowerMonitor{}, errReadShunt
+		err = errReadShunt
+		return
 	}
 	// Least significant bit is 10µV.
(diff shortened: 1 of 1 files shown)


maruel (approved):
Thanks!

PR #88, opened 2025-01-13
Add Support for SparkFun SerLCD Display

gsexton:
Bumps conn to 3.7.2 to get the new text interface, and the GPIO group interface.

### DIFF
--- a/serlcd/serlcd.go
+++ b/serlcd/serlcd.go
@@ -1,0 +1,249 @@
+// Copyright 2024 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// This package provides an implementation for the SparkFun SerLCD intelligent
+// LCD display. This display provides hardware interfaces for SPI, I2C, and
+// UART. Implements conn.display.TextDisplay
+package serlcd
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	"periph.io/x/conn/v3"
+	"periph.io/x/conn/v3/display"
+)
+
+// Representation of a SerLCD display.
+type Dev struct {
+	conn conn.Conn
+	w    io.Writer
+	cols int
+	rows int
+	// Display on/off, Curosr, Blink
+	displayDCB byte
+}
+
+const (
+	DefaultI2CAddress uint16 = 0x72
+)
+
+var settingMode byte = 0x7c
+var cmdMode byte = 0xfe
+var clear = []byte{settingMode, 0x2d}
+
+func wrap(err error) error {
+	return fmt.Errorf("serlcd: %w", err)
+}
+
+// Create a SerLCD display using a hardware interface that provides io.Writer.
+// That can be the i2c.Bus, or a 3rd Party UART library for serial
+// communications.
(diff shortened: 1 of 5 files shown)


serlcd/serlcd.go
```diff
@@ -0,0 +1,249 @@
...
+var clear = []byte{settingMode, 0x2d}
+
+func wrap(err error) error {
+	return fmt.Errorf("serlcd: %w", err)
+}
+
+// Create a SerLCD display using a hardware interface that provides io.Writer.
+// That can be the i2c.Bus, or a 3rd Party UART library for serial
+// communications.
+func NewSerLCD(writer io.Writer, rows, cols int) *Dev {
```
maruel, line 45: Use `New`, as `serlcd.NewSerLCD` stutters. Maybe `NewSerial`?

serlcd/serlcd.go
```diff
@@ -0,0 +1,249 @@
...
+	dev := &Dev{w: writer, rows: rows, cols: cols}
+	_ = dev.Display(true)
+	return dev
+}
+
+// Create a SerLCD display using a hardware interface that provides
+// conn.Conn. For example, a conn.spi.Conn
+func NewConnSerLCD(conn conn.Conn, rows, cols int) *Dev {
```
maruel, line 53: `NewConn`

serlcd/serlcd.go
```diff
@@ -0,0 +1,249 @@
...
+	return fmt.Sprintf("SparkFun SerLCD %dx%d Display - %s", dev.cols, dev.rows, ioType)
+}
+
+// Write a set of bytes to the display.
+func (dev *Dev) Write(p []byte) (n int, err error) {
+	if dev.w != nil {
+		n, err = dev.w.Write(p)
+		return
+	}
+	// Evidently, for i2c there's a buffer limitation of 32 bytes. Writing
```
maruel, line 200: Isn't that device specific? It's okay here but please mention "Device XYZ has a write buffer of 32 bytes. Make it configurable in the future if needed."

maruel (approved):
lgtm with a few changes

serlcd/serlcd.go
gsexton: The proper name of the product is SerLCD

serlcd/serlcd.go
gsexton: This is a really specific device. They've got a board with an Atmel processor on it running a custom arduino image and that talks to the display, and drives the LEDs for the RGB. There's no way to make this more generic than having the interface that it complies with.

maruel: An alternative is to have the New**() functions to accept an Options struct, so that this could become configurable in the future as a non-breaking change? For now, the Options struct would contain cols and rows.

gsexton: I've done a little work on this. I've included a link to the SparkFun issue about the 32 byte limit. The issue is 5 years old, and doesn't have a resolution and from the tone, it never will. 

This isn't an actively developed product category, so I think adding complexity for future compatibility is just adding complexity. I've restructured the 32 char limit to make it more transparent what's going on should it change in the future.

maruel (approved):
lgtm

PR #91, opened 2025-02-07
hd44780: Restructure to implement conn/display/TextDisplay. Refine to use GPIO.Group

### LINKED ISSUE #3: I2C IO Expander for HD44780
What would be the best way to go about implementing a driver for the hd44780 in 4/8 bit mode over an 8-bit I2C IO Expander like the [PCF8574T](http://wiki.sunfounder.cc/images/1/18/PCF8574T_datasheet.pdf) as done [for arduino in this project](https://gitlab.com/joearmstrong980/LCD_I2C), [for Raspberry pi](https://github.com/sunfounder/SunFounder_SensorKit_for_RPi2/blob/master/Python/30_i2c_lcd1602.py), and in go [here with gobot](https://github.com/ChrisTrenkamp/gobotlcd), and standalone [here](https://github.com/d2r2/go-hd44780) and [here](https://github.com/wjessop/lcm1602_lcd)?

After reviewing the Project Design and Guidelines, it's not completly clear to me what the intended way to do this is.

Some ways I can think of would be:
- making a new driver that uses the hd44780 driver code but has code to use the IO Expander instead of a direct pin connection, 
- changing the existing driver to allow for both direct usage and usage over I2C,
- or maybe to implement I2C IO Expanders in the conn/i2c module.

Currently, I'm using gobot to drive my LCD, though this adds redundancy as both periph and [gobot](gobot.io) are packages for the same or at least similar purposes and i would like to keep my dependencies to a minimum to keep a small overhead in my project.

gsexton:
Fixes #3


hd44780/af_i2c_backpack.go
```diff
@@ -0,0 +1,61 @@
...
+	mcp, err := mcp23xxx.NewI2C(bus, mcp23xxx.MCP23008, address)
+	if err != nil {
+		return nil, err
+	}
+	gr := *mcp.Group(0, []int{d4, d5, d6, d7, rsPin, enablePin, backlightPin})
+	grPins := gr.Pins()
+	reset := grPins[4].(gpio.PinOut)
+	enable := grPins[5].(gpio.PinOut)
+	bl := grPins[6].(gpio.PinOut)
+	return NewHD44780(gr, &reset, &enable, &bl, rows, cols)
```
maruel, line 47: I believe you don't want to take the address of the local variables.

maruel (approved):
lgtm once it passes checks

gsexton:
@maruel Did you have a solution for the pointer issue you raised? 

Also, to get this in requires the updated version of host. Should I do a release of host, and then update devices to use the new release? Is there better way to handle this?

hd44780/hd44780.go
```diff
@@ -14,216 +14,427 @@ import (
...
+// The first 4 or 8 pins of the data group must be connected to the data lines
+// To use 4 bit mode, you would connect lines D4-D7 on the display, and for 8
+// bit mode, D0-D7. If dataPinGroup is 8 or more pins, then it's assumed the
+// display is connected using all 8 pins.
+//
+// backlight should implement either display.DisplayBacklight or
+// display.DisplayRGBBacklight. See GPIOMonoBacklight.
+func NewHD44780(
+	dataPinGroup gpio.Group,
+	resetPin, enablePin *gpio.PinOut,
```
maruel, line 86: gpio.PinOut is an interface, so you do not want to take the pointer, it is already a pointer.

maruel:
> @maruel Did you have a solution for the pointer issue you raised?

Oh it's a mistake between an interface and a pointer. In Go, an interface is itself a pointer.

> Also, to get this in requires the updated version of host. Should I do a release of host, and then update devices to use the new release? Is there better way to handle this?

Yes, that's the current way to do it, in two steps. Can you do it? Thanks!

PR #92, opened 2025-02-09
mcp23xxx: Add GPIO Group support

### LINKED ISSUE #17: Direct Register Access for MCP23xxx

I'm trying to figure out how to get direct access to the MCP23008 GPIO register using the MCP23xxx package. The only interface I see exposed is via pins. I'm writing a driver for the Adafruit I2c/SPI backpack, and it uses the MCP23008.

The need is because to send a character to the backpack you have to:

Set the reset pin on the GPIO
Write A value to pins 3-6 (data), and pin 2 low (enable)
Toggle Pin 2 High
Wait a period
Toggle Pin 2 Low
Write A value to pins 3-6 (data), and pin 2 low (enable)
Toggle Pin 2 High
Wait a period
Toggle Pin 2 Low

If I have to set each pin individually, that stacks up to a fair number of I2c transactions.

- Hardware: MCP23008 (on Adafruit i2c/SPI backpack)
- Software: Linux, Raspbian, I2C

**Do you plan to:**
If there's not a method, I could contribute a patch.

- Contribute an initial driver: I could contribute a patch.
- Write unit tests: Yes
- Update https://github.com/periph/cmd to use the new functionality: No

Thanks

George

gsexton:
Fixes #17


maruel:
Do I understand you need to rebase first so the tests pass?

mcp23xxx/group.go
```diff
@@ -0,0 +1,214 @@
...
+	var pgif interface{} = &pinGroup{dev: dev, port: port, pins: grouppins, defaultMask: defMask}
+	if gpiogroup, ok := pgif.(gpio.Group); ok {
+		return &gpiogroup
+	}
+
+	return nil
+}
+
+// Pins returns the set of pin.Pin that make up that group.
+func (pg *pinGroup) Pins() []pin.Pin {
```
maruel, line 44: One thing I noticed: since gpio.Group is about gpio.Pin, should the methods of the interfaces return gpio.PinIO instead pin.Pin? This would remove a lot of casting for the end user.

maruel (approved):
lgtm once tests pass.

mcp23xxx/group.go
gsexton: PinIO is both read and write. There are some devices that don't support read.

maruel: Yeah I understand the trade off. It's okay as is.

maruel (approved):

PR #95, opened 2025-02-14
inky: run stringer with "go run"

dolmen:
Run the `stringer` command (for code generation) using `go run` instead of `go install` to avoid affecting the developer's global `$GOBIN`.

### DIFF
--- a/inky/types.go
+++ b/inky/types.go
@@ -4,7 +4,7 @@
 
 package inky
 
-//go:generate go install golang.org/x/tools/cmd/stringer@latest
+//go:generate -command stringer go run golang.org/x/tools/cmd/stringer@latest
 //go:generate stringer -type=Model,Color,ImpressionColor -output types_string.go
 
 import (


inky/types.go
```diff
@@ -4,7 +4,7 @@
 
 package inky
 
-//go:generate go install golang.org/x/tools/cmd/stringer@latest
+//go:generate -command stringer go run golang.org/x/tools/cmd/stringer@latest
 //go:generate stringer -type=Model,Color,ImpressionColor -output types_string.go
```
maruel, line 8: Thanks, #til! Did you mean to merge the two lines?

maruel (changes requested):

inky/types.go
dolmen: I think that keeping the two lines allows to avoid a very long line where stringer arguments would be on the far right. This helps readability.

dolmen: @maruel ^^^ Do you still want me to merge the two lines?

dolmen:
Note that Go 1.24.0 introduces [special support for version-locking of build tools](https://go.dev/doc/modules/managing-dependencies#tools) such as `stringer` in go.mod. With that support the single line would be:
```go
//go:generate go tool stringer ...
```

Of course it is a bit early to require `go 1.24.0` in go.mod. I'm just mentionning this for a not-so-near future upgrade.

inky/types.go
maruel: Ah ok I understand now.

maruel (approved):
Thanks!

PR #96, opened 2025-02-14
lepton/cci: run stringer with "go run"

dolmen:
Run the `stringer` command (for code generation) using `go run` instead of `go install` to avoid affecting the developer's global `$GOBIN`.

See also #95.


maruel (approved):
Thanks!

PR #99, opened 2025-02-23
Add Support for 74HC595 Serial -> Parallel Shift Register

gsexton:
The 74HC595 converts a serial data stream to parallel output on it's data pins. The Adafruit I2C/SPI backpack uses this device for the SPI side of the backpack.

I'll be adding an SPI backpack to the HD44780 LCD display that uses this.

### DIFF
--- a/nxp74hc595/74hc595.go
+++ b/nxp74hc595/74hc595.go
@@ -1,0 +1,190 @@
+// Copyright 2025 The Periph Authors. All rights reserved.
+// Use of this source code is governed under the Apache License, Version 2.0
+// that can be found in the LICENSE file.
+
+// The 74HC595 is a serial shift register. It converts a serial stream to a
+// parallel output. For example, you can use it as an SPI => Parallel
+// converter.
+//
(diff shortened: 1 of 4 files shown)


nxp74hc595/74hc595.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// WaitForEdge is not available for this device.
+func (gr *Group) WaitForEdge(timeout time.Duration) (int, gpio.Edge, error) {
+	return 0, gpio.NoEdge, ErrNotImplemented
+}
+
+// Halt frees the group's resources and prevents it from being used again.
+func (gr *Group) Halt() error {
+	gr.pins = make([]Pin, 0)
```
maruel, line 176: gr.pins = nil?

maruel (approved):
Thanks!

nxp74hc595/74hc595.go
gsexton: Sure.

Issue #23, opened 2021-11-05
Support MPU9250 over I2C


bryanjhv:
**What kind of new feature are you looking for?**
- Hardware: Support MPU9250 over I2C, currently only implementation uses SPI which is fine but many mixed chips like the GY-91 expose it as I2C. Another option would be to make `mpu9250.Proto` interface with public methods so one can provide an implementation for it (like `SpiTransport` does).

**Do you plan to:**
- Contribute an initial driver: No
- Write unit tests: No
- Update https://github.com/periph/cmd to use the new functionality: No

bezineb5:
I have an implementation of MPU9250 over I2C, with embedded AK8963 I2C magnetometer. However, it's not based upon the code of the existing MPU9250 driver, so I never submitted it (the interfaces are not compatible).
I can share the code if still needed.
@maruel Any interest in this MPU9250 I2C + AK8963? (the AK8963 is actually the magnetometer part of the MPU9250, exposed as a separate device on the I2C bus)

maruel:
Sure!

gsexton:
Implemented in 

https://github.com/periph/devices/pull/101

PR #101, opened 2025-03-15
mpu9250: Add i2c interface support

gsexton:
Implements #23


mpu9250/transport.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"periph.io/x/conn/v3/gpio"
+	"periph.io/x/conn/v3/i2c"
+	"periph.io/x/conn/v3/physic"
+	"periph.io/x/conn/v3/spi"
+	"periph.io/x/conn/v3/spi/spireg"
+)
+
+// DebugF the debug function type.
+type DebugF func(string, ...interface{})
```
maruel, line 18: Side note: we should probably standardize on the log/slog package. It's actually pretty good.

maruel (approved):

PR #104, opened 2025-03-28
ssd1306: Add support for SH1106 and SH1107

gsexton:
Add support for SH1106, and SH1107 displays.
Expand example with some drawing primitives.

### DIFF
--- a/ssd1306/image1bit/image1bit.go
+++ b/ssd1306/image1bit/image1bit.go
@@ -144,17 +144,17 @@ func (i *VerticalLSB) SetBit(x, y int, b Bit) {
 }
 
 // Draw a horizontal line from start to end at ypos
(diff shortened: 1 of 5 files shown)


ssd1306/image1bit/image1bit.go
```diff
@@ -143,6 +143,20 @@ func (i *VerticalLSB) SetBit(x, y int, b Bit) {
 	}
 }
 
+// Draw a horizontal line from start to end at ypos
+func (i *VerticalLSB)DrawHLine(start, end, ypos int, b Bit) {
```
maruel, line 147: You may want to configure your editor to format on save. This is quite useful as then you can type without caring about formatting and a quick file save formats for you.

maruel (approved):
lgtm after linter/test pass

PR #106, opened 2025-04-04
inky: Fix Impression Driver

### LINKED ISSUE #105: Inky Impression 7.3 Does not Work As Expcted
From @kedare

```
Hmm I could not make it work so far, I found a few issues so far:
Dev.Bounds() is always (0,0),(0,0), it's not properly set as it should automatically depending on the model (I had to force is via Opts.Height and Opts.Width)

    When Drawing, it would randomly fail with the following logs :

2025/03/02 19:22:47 impression.go:558: Err: bcm283x-gpio (GPIO17): line_request ioctl: bad file descriptor
2025/03/02 19:22:48 impression.go:558: Err: bcm283x-gpio (GPIO17): line_request ioctl: bad file descriptor
2025/03/02 19:22:48 impression.go:558: Err: bcm283x-gpio (GPIO17): line_request ioctl: bad file descriptor
2025/03/02 19:22:48 impression.go:558: Err: bcm283x-gpio (GPIO17): line_request ioctl: bad file descriptor

Which apparently correspond to this : https://github.com/pimoroni/inky/issues/205
But on the python implementation this is not making the drawing fail, those are just warnings, it looks like on the Go implementation this stops the drawing (can't confirm as I can't draw at all)

    When not raising the previous error, it's not actually drawing anything, it does block however the code like if it was indeed drawing something (around 40s)

This is the code I am using: https://github.com/kedare/dashink/blob/main/pkg/hardware/host.go

Thanks
```

**To Reproduce**
1. Modify the cmd/inky program call inky.NewImpression()
2. Build the program and call with an appropriate image.

Output:

2025/03/02 19:22:47 impression.go:558: Err: bcm283x-gpio (GPIO17): line_request ioctl: bad file descriptor

in my instrumented version, I get:

inky: GPIOLine(27).Out(): line_request chip_fd=7, ioctl: bad file descriptor.

**Expected behavior**
The program should render the image on the Display

 - OS: Raspbian Bookworm
 - Board Raspbery Pi 5 w/ 8GB
 - Device Inky Impression 7.3

**Additional context**

I changed the program to use the gpioioctl package directly since the impression file depends upon WaitForEdge(). The issue is coming in this function:

```go
// Render renders the content of the [*DevImpression.Pix] to the screen.
func (d *DevImpression) Render() error {
	fmt.Println("In Render() Press Enter to begin render calls.")
	fmt.Scanln()
	if d.flipVertically {
		for w := 0; w < len(d.Pix)/2-1; w = w + d.width {
			for offset := 0; offset < d.width; offset++ {
				d.Pix[w+offset], d.Pix[len(d.Pix)-d.width-w+offset] = d.Pix[len(d.Pix)-d.width-w+offset], d.Pix[w+offset]
			}
		}
	}

	if d.flipHorizontally {
		for offset := 0; offset < len(d.Pix)-1; offset = offset + d.width {
			for i, j := 0, d.width-1; i < j; i, j = i+1, j-1 {
				d.Pix[i+offset], d.Pix[j+offset] = d.Pix[j+offset], d.Pix[i+offset]
			}
		}
	}
	log.Println("Preparing to create Merged(). Press Enter to Begin")
	fmt.Scanln()
	merged := make([]uint8, len(d.Pix)/2)

	log.Printf("len merged=%d\n Press Enter to continue", len(merged))

	fmt.Scanln()
	for i, offset := 0, 0; i < len(d.Pix)-1; i, offset = i+2, offset+1 {
		merged[offset] = (d.Pix[i]<<4)&0xF0 | d.Pix[i+1]&0x0F
		log.Printf("offset=%d", offset)
	}
	log.Println("Merged created.")
	log.Println("In Render() Press Enter to call Update()")
	fmt.Scanln()
	return d.update(merged)
}
```
When I run the Linux command lsof -p pid at the step Preparing to create Merged(). Press Enter to Begin, I see the following:

```bash
gsexton@raspberrypi:/proc/7019/fd $ ps xfa | grep main
  14698 pts/0    Sl+    0:00  |   \_ ./main -image /home/gsexton/George.png

gsexton@raspberrypi:/proc/7019/fd $ lsof -p 14698
COMMAND   PID    USER   FD      TYPE DEVICE SIZE/OFF   NODE NAME
main    14698 gsexton  cwd       DIR  179,2     4096 649617 /home/gsexton/periph/x/cmd/inky
main    14698 gsexton  rtd       DIR  179,2     4096      2 /
main    14698 gsexton  txt       REG  179,2  4169165 657764 /home/gsexton/periph/x/cmd/inky/main
main    14698 gsexton    0u      CHR  136,0      0t0      3 /dev/pts/0
main    14698 gsexton    1u      CHR  136,0      0t0      3 /dev/pts/0
main    14698 gsexton    2u      CHR  136,0      0t0      3 /dev/pts/0
main    14698 gsexton    3r      REG  179,2    20475  95166 /home/gsexton/George.png
main    14698 gsexton    4u  a_inode   0,14        0   1183 [eventpoll:5,6,7,8,9,10,11,12]
main    14698 gsexton    5u  a_inode   0,14        0   1183 [eventfd:191]
main    14698 gsexton    6r      CHR  254,0      0t0    132 /dev/gpiochip0
main    14698 gsexton    7r      CHR 254,10      0t0     88 /dev/gpiochip10
main    14698 gsexton    8r      CHR 254,11      0t0     89 /dev/gpiochip11
main    14698 gsexton    9r      CHR 254,12      0t0     90 /dev/gpiochip12
main    14698 gsexton   10r      CHR 254,13      0t0     91 /dev/gpiochip13
main    14698 gsexton   11r      CHR  254,0      0t0    132 /dev/gpiochip0
main    14698 gsexton   12w      REG   0,18    16384   3709 /sys/class/gpio/export
main    14698 gsexton   13u      CHR  153,1      0t0    665 /dev/spidev0.0

```

when I press enter, and this is executed:

```go
	merged := make([]uint8, len(d.Pix)/2)
	log.Printf("len merged=%d\n Press Enter to continue", len(merged))
	fmt.Scanln()
```
the program output is:

```
2025/03/29 14:28:26 len merged=192000
 Press Enter to continue
```

running lsof -p again shows the expected gpiochip file handles open. Pressing ENTER to execute the code:

```go
	for i, offset := 0, 0; i < len(d.Pix)-1; i, offset = i+2, offset+1 {
		merged[offset] = (d.Pix[i]<<4)&0xF0 | d.Pix[i+1]&0x0F
		log.Printf("offset=%d", offset)
	}
	log.Println("Merged created.")
	log.Println("In Render() Press Enter to call Update()")
	fmt.Scanln()
```
the output is:

```
2025/03/29 14:54:05 offset=191999
2025/03/29 14:54:05 Merged created.
```

and then re-running lsof -p shows:

```bash
gsexton@raspberrypi:/proc/7019/fd $ lsof -p 14698
COMMAND   PID    USER   FD      TYPE DEVICE SIZE/OFF   NODE NAME
main    14698 gsexton  cwd       DIR  179,2     4096 649617 /home/gsexton/periph/x/cmd/inky
main    14698 gsexton  rtd       DIR  179,2     4096      2 /
main    14698 gsexton  txt       REG  179,2  4169165 657764 /home/gsexton/periph/x/cmd/inky/main
main    14698 gsexton    0u      CHR  136,0      0t0      3 /dev/pts/0
main    14698 gsexton    1u      CHR  136,0      0t0      3 /dev/pts/0
main    14698 gsexton    2u      CHR  136,0      0t0      3 /dev/pts/0
main    14698 gsexton    3r      REG  179,2    20475  95166 /home/gsexton/George.png
main    14698 gsexton    4u  a_inode   0,14        0   1183 [eventpoll:5,12]
main    14698 gsexton    5u  a_inode   0,14        0   1183 [eventfd:191]
main    14698 gsexton   12w      REG   0,18    16384   3709 /sys/class/gpio/export
main    14698 gsexton   13u      CHR  153,1      0t0    665 /dev/spidev0.0

```
Note that the file descriptors to /dev/gpiochip* are gone. The error is being returned by cycleResetGPIO() because the Out() command is attempting to use the gpiochip0 file descriptor to perform an ioctl() for the pin:

```go
func (d *DevImpression) cycleResetGPIO() error {
	if err := d.r.Out(gpio.Low); err != nil {
		return err
	}
	time.Sleep(100 * time.Millisecond)
	return d.r.Out(gpio.High)
}
```
The behavior isn't perfectly consistent. Sometimes creating the slice is enough to change the output of lsof, sometimes it isn't until the loop is done. It is  not 100% consistent that it happens, but when it does, it's in one of those two spots.

The mainImpl() function has a reference to a gpioioctl.Chip[0],

I instrumented the GPIO chip to print when Close() is called, and that's not happening.

Also, I ran it under strace and don't see close calls with any of the affected file descriptors. On ArmV6, the ioctl error does not occur, but it just doesn't work. There are probably multiple issues.

@maruel Can you look at this? I'm stumped. It seems like either some sort of memory overwrite is happening, or something is garbage collecting things. I confirmed that the length of gpioioctl.Chips[0] is still well-constructed after the error happens, so it doesn't look like Init() is called twice making things strange.

gsexton:
Fixes #105

This PR fixes issues in the Inky 7.3 Impression Driver.

* Changed the SPI CS pin to work under automatic or manual control. This allows it to work when CS is controlled by SPI, or when it is controlled manually. We don't want it to fail in the configuration the Pimoroni python driver requires.
* Changed the wait() routine to use a 10ms debounce on WaitForEdge() as the python driver does. 
* Fixed an issue in the image copy where the image byte wasn't copied (see line 530 of impression.go).
* If the BUSY line is high on entry to wait to go directly to a timed wait as the python driver does.
* Fixed errors in the palette definitions. Removed extra memory allocation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/mattn/go-colorable v0.1.13
 	golang.org/x/image v0.23.0
 	periph.io/x/conn/v3 v3.7.2
-	periph.io/x/host/v3 v3.8.4
+	periph.io/x/host/v3 v3.8.5
 )
 
 require (
--- a/go.sum
+++ b/go.sum
@@ -17,5 +17,5 @@ golang.org/x/sys v0.29.0 h1:TPYlXGxvx1MGTn2GiZDhnjPA9wZzZeGKHHmKhHYvgaU=
 golang.org/x/sys v0.29.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 periph.io/x/conn/v3 v3.7.2 h1:qt9dE6XGP5ljbFnCKRJ9OOCoiOyBGlw7JZgoi72zZ1s=
 periph.io/x/conn/v3 v3.7.2/go.mod h1:Ao0b4sFRo4QOx6c1tROJU1fLJN1hUIYggjOrkIVnpGg=
-periph.io/x/host/v3 v3.8.4 h1:QNleTythDd0k6Chu0n+ISrJFlf3LFig9oNbtOIkxoCc=
-periph.io/x/host/v3 v3.8.4/go.mod h1:hPq8dISZIc+UNfWoRj+bPH3XEBQqJPdFdx218W92mdc=
+periph.io/x/host/v3 v3.8.5 h1:g4g5xE1XZtDiGl1UAJaUur1aT7uNiFLMkyMEiZ7IHII=
+periph.io/x/host/v3 v3.8.5/go.mod h1:hPq8dISZIc+UNfWoRj+bPH3XEBQqJPdFdx218W92mdc=
--- a/inky/impression.go
+++ b/inky/impression.go
@@ -17,7 +17,6 @@ import (
 	"periph.io/x/conn/v3"
 	"periph.io/x/conn/v3/display"
 	"periph.io/x/conn/v3/gpio"
-	"periph.io/x/conn/v3/gpio/gpioreg"
 	"periph.io/x/conn/v3/physic"
 	"periph.io/x/conn/v3/spi"
 )
@@ -160,7 +159,7 @@ func NewImpression(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinI
 	// is 192K Bytes. To make the Impression 7.3 treat this as single trans-
 	// action, we have to take over control of the CS pin and manipulate it
 	// as required.
-	c, err := p.Connect(cSpeed, spi.Mode0|spi.NoCS, 8)
+	c, err := p.Connect(cSpeed, spi.Mode0, 8)
 	if err != nil {
 		return nil, fmt.Errorf("failed to connect to inky over spi: %v", err)
 	}
@@ -191,7 +190,6 @@ func NewImpression(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinI
 			model:      o.Model,
 			variant:    o.DisplayVariant,
 			pcbVariant: o.PCBVariant,
-			cs:         gpioreg.ByName(cs0Pin),
 		},
 		saturation: 50, // Looks good enough for most of the images.
 	}
--- a/inky/impression.go
+++ b/inky/impression.go
@@ -156,10 +156,6 @@ func NewImpression(p spi.Port, dc gpio.PinOut, reset gpio.PinOut, busy gpio.PinI
 	if o.Model == IMPRESSION73 {
 		cSpeed = acSpeed
 	}
-	// The SPI driver has a max buffer size of 4K bytes, but our image size
-	// is 192K Bytes. To make the Impression 7.3 treat this as single trans-
-	// action, we have to take over control of the CS pin and manipulate it
-	// as required.
 	c, err := p.Connect(cSpeed, spi.Mode0, 8)
 	if err != nil {
 		return nil, fmt.Errorf("failed to connect to inky over spi: %v", err)
--- a/inky/impression.go
+++ b/inky/impression.go
@@ -29,36 +29,36 @@ var _ draw.Image = &DevImpression{}
 var (
 	// For more: https://github.com/pimoroni/inky/issues/115#issuecomment-887453065
 	dsc = []color.NRGBA{
-		{0, 0, 0, 0},         // Black
+		{0, 0, 0, 255},       // Black
 		{255, 255, 255, 255}, // White
 		{0, 255, 0, 255},     // Green
 		{0, 0, 255, 255},     // Blue
 		{255, 0, 0, 255},     // Red
 		{255, 255, 0, 255},   // Yellow
 		{255, 140, 0, 255},   // Orange
-		{255, 255, 255, 255},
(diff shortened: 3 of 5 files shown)


gsexton:
Rats! I'm hitting an issue. On a Pi 5, when I call:

c, err := p.Connect(cSpeed, spi.Mode0|spi.NoCS, 8)

I get:

**inky: failed to connect to inky over spi: sysfs-spi: setting mode Mode0|NoCS failed: invalid argument.**

I've tried both with and without

dtoverlay=spi0-0cs 

in /boot/firmware/config.txt

@maruel  do you have any ideas? I suppose I can look at how to configure SPI for an unused GPIO pin as CS pin.

gsexton:
Still thinking through this a little. I've got a some more testing to do.

gsexton:
OK, this is ready now.

edwinavalos:
Great timing! I just tried to use the `periph` inky libraries for the first time yesterday and was running into some of the issues outlined in #105. Big difference in my attempts is that I am trying to control a inky WHAT yellow. 

I pulled down this branch and was able to get my display updating correctly now instead of the `cmd/inky` command exiting with no updates. Thank you for taking the time to fix this, a random stranger thanks you!

maruel (approved):

PR #107, opened 2025-05-05
Support Inky Impression 7.3 7-color

briancharous:
This PR adds support for the Inky Impression 7.3 7-color model. The only relevant change is mapping the PCB variant to the device model when trying to detect the device's opts.


maruel (approved):
Thanks!

PR #109, opened 2025-05-27
Introduces driver for Sensirion SHT-4X Temperature/Humidity Sensors

gsexton:
Also moves CRC8 logic into common package for re-usability across multiple packages.


maruel (approved):
Thanks!

I'm not super enthusiast with the package name "common" but I don't have a good suggestion readily available so it's fine.

